PR #11, opened 2012-11-10
Making cppzmq cmake and submodule friendly

smadhueagle:


kobolog:
I'm gonna close it since it's not mergeable and wasn't touched for a while.

PR #42, opened 2015-03-05
Fix a misaligned load

tolchz:
The issue was discovered by clang:

zmq.hpp:504:34: runtime error: load of misaligned address 0x2b38beb224b2
for type 'int32_t' (aka 'int'), which requires 4 byte alignment

I modified the loads of both event and value to be consistent.

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -500,8 +500,8 @@ namespace zmq
 #if ZMQ_VERSION_MAJOR >= 4
                 const char* data = static_cast<const char*>(zmq_msg_data(&eventMsg));
                 zmq_event_t msgEvent;
-                msgEvent.event = *(uint16_t*)data; data += sizeof(uint16_t);
-                msgEvent.value = *(int32_t*)data;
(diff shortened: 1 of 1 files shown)


zmq.hpp
```diff
@@ -500,8 +500,8 @@ namespace zmq
 #if ZMQ_VERSION_MAJOR >= 4
                 const char* data = static_cast<const char*>(zmq_msg_data(&eventMsg));
                 zmq_event_t msgEvent;
-                msgEvent.event = *(uint16_t*)data; data += sizeof(uint16_t);
-                msgEvent.value = *(int32_t*)data;
+                memcpy(&msgEvent.event, data, sizeof(int16_t)); data += sizeof(uint16_t);
```
Kobolog: Shouldn't it be `uint16_t` in both cases?

tolchz: Yup, typo, good catch.

Kobolog:
Thanks!

PR #43, opened 2015-03-05
Add an rvalue overload for socket_t::send

tolchz:
This lets you create functions that return message_t's by value, and
pass them to send() with no copy.

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -441,6 +441,13 @@ namespace zmq
             throw error_t ();
         }
 
+#ifdef ZMQ_HAS_RVALUE_REFS
+        inline bool send (message_t &&msg_, int flags_ = 0)
+        {
+            return send(msg_, flags_);
+        }
+#endif
+
         inline size_t recv (void *buf_, size_t len_, int flags_ = 0)
         {
             int nbytes = zmq_recv (ptr, buf_, len_, flags_);


zmq.hpp
```diff
@@ -441,6 +441,13 @@ namespace zmq
             throw error_t ();
         }
 
+      #ifdef ZMQ_HAS_RVALUE_REFS
```
Kobolog: The preprocessor lines are not indented throughout the file.

zmq.hpp
```diff
@@ -441,6 +441,13 @@ namespace zmq
...
+        inline bool send (message_t &&msg_, int flags_ = 0)
+        {
+          return send(msg_, flags_);
```
Kobolog: And the indentation here is off as well.

tolchz: Got it, thanks.

zmq.hpp
```diff
@@ -441,6 +441,13 @@ namespace zmq
             throw error_t ();
         }
 
+#ifdef ZMQ_HAS_RVALUE_REFS
+        inline bool send (message_t &&msg_, int flags_ = 0)
+        {
+          return send(msg_, flags_);
```
Kobolog: Still two space indentation here unlike four in other functions.

tolchz: Think this has it resolved.

Kobolog:
Yes, thank you.

PR #44, opened 2015-03-06
Don't throw exceptions when using ZMQ_ROUTER_MANDATORY

tolchz:
Previously when using the ZMQ_ROUTER_MANDATORY option on a socket and
sending to an unknown destination, an exception was thrown.

This change makes the send functions to treat the EAGAIN and
EHOSTUNREACH errors similarly.  The user can check zmq_errno() to tell
them apart.

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -426,7 +426,8 @@ namespace zmq
             int nbytes = zmq_send (ptr, buf_, len_, flags_);
             if (nbytes >= 0)
                 return (size_t) nbytes;
-            if (zmq_errno () == EAGAIN)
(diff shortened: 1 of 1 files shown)


tolchz:
I'm not entirely sure if this is the best way to handle this.  To me it seems like an unreachable host shouldn't warrant an exception.

Kobolog:
It might be a good idea to change that, but it's an API breaking change. We might break a lot of code relying on a fact that an unreachable destination will generate an exception without actually checking the return value.

tolchz:
No problem, I'll close this one.

Issue #58, opened 2015-11-11
Visual studio 2013 building (another one)


brucelane:
Hello, 
I get this error:
error C2664: 'int zmq_msg_recv(zmq_msg_t *,void *,int)' : cannot convert 'void *' in 'zmq_msg_t *'	t:\zeromq\cppzmq\zmq.hpp	666	1	cpp.hwclient

thank you for any hint

brucelane:
should it be zmq_msg_recv (&addrMsg, s, 0) instead of zmq_msg_recv (s, &addrMsg, 0) ?
do I have to compile as 32bit or 64bit?

imnotfromaustralia:
Hello! I have the same problem too. There's different function signatures:
`zmq_msg_recv (zmq_msg_t *msg, void *s, int flags)`
vs
`zmq_recvmsg (void *s, zmq_msg_t *msg, int flags)`
so I think the solution is to rewrite all func calls to zmq_recvmsg to match the signature

xaqq:
Since a commit addressing the issue has been merge I assume we can close the issue.
Please feel free to reopen if needed.

PR #59, opened 2015-11-12
Fix usage of zmq_msg_recv and zmq_recvmsg

dquam:
The zmq_msg_recv() function takes its parameters in the order of message, socket, and flags while the zmq_recvmsg() function takes the socket, message, and flags. This commit addresses that difference.


dquam:
See issue #58

PR #60, opened 2015-11-13
fixed error introduced by previous commit

cmcmurrough:
Previous commit replaced deprecated calls with zmq_msg_recv, but the first and second arguments were in the wrong order.

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -663,7 +663,7 @@ namespace zmq
             while (true) {
                 zmq_msg_t eventMsg;
                 zmq_msg_init (&eventMsg);
-                rc = zmq_msg_recv (s, &eventMsg, 0);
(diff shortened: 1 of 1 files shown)


c-rack:
Thanks, but it seems to be a dupe of #59, right?

dquam:
If you expand [the macro on line 91](https://github.com/cmcmurrough/cppzmq/blob/master/zmq.hpp#L91), you will get build errors because the parameters are again out of order right?

cmcmurrough:
Sorry, I didn't notice #59 when I made the request. Dquam's should be merged, as his commit address the macro thing. Sorry for the dupe, I will close this now.

c-rack:
Just merged #59.
Thanks @cmcmurrough and @dquam for your contributions!

PR #68, opened 2015-12-03
Rename message_t to frame_t

K0n63n:
Makes naming more consistent with the CZMQ high-level C binding. Furthermore, frees the message_t name, such that it can be reused to implement a multipart message corresponding to the zmsg_t of CZMQ.

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -189,27 +189,27 @@ namespace zmq
     }
     #endif
 
-    class message_t
+    class frame_t
     {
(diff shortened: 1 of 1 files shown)


ricnewton:
Won't this break a whole ton of user code that is out there?

K0n63n:
Yes, you're right. It's probably best to leave it as it is.

ricnewton:
Thanks.

Issue #74, opened 2016-03-23
assert error in the context_t


pengweichu:
I think in the content_t class, we must check the ptr has been destroyed or not, otherwise if we called the close before destroy, then the ZMQ_ASSERT will be raise when the class destroying.

Below is the modified code:

```
inline ~context_t () ZMQ_NOTHROW
{
    if (!ptr)
    {
        return;
    }

    int rc = zmq_ctx_destroy (ptr);
    ptr = NULL;
    ZMQ_ASSERT (rc == 0);
}

inline void close() ZMQ_NOTHROW
{
    if (!ptr)
    {
        return;
    }

    int rc = zmq_ctx_destroy (ptr);
    ptr = NULL;
    ZMQ_ASSERT (rc == 0);
}
```

ignitusboyone:
According to the man page

http://api.zeromq.org/3-2:zmq-ctx-destroy
http://api.zeromq.org/4-2:zmq-ctx-destroy

The zmq_ctx_destroy() function shall return zero if successful. Otherwise it shall return -1 and set errno to one of the values defined below

So why are we asserting that close and ~context_t were successful.  Shouldn't the assert check should be  rc != 0 or rc == -1 ?

sigiesec:
This issue was fixed by #85

PR #78, opened 2016-06-29
Fixed context_t::close and context_t::~context_t

sigiesec:
Restored behaviour before 014628cb34735cfdf8d049e2316574f713361e5a and 971092e472b5b68d36334e88cb35f75972c35415 such that the context is properly destroyed AND close may be called explicitly without causing an assertion in the dtor

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -410,13 +410,16 @@ namespace zmq
 
         inline ~context_t () ZMQ_NOTHROW
         {
-            int rc = zmq_ctx_destroy (ptr);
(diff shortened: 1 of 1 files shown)


sigiesec:
This should fix #74

c-rack:
@sigiesec could you please resolve the conflict in order to make your PR mergable?

sigiesec:
This overlaps with #85, which chose a bit different solution that also allows close() to be called multiple times. I don't think that that should be allowed, but I am not familiar with the overall design conventions of cppzmq, so I am closing this pull request.

PR #82, opened 2016-08-14
Enable passing flags in to `send()` and `recv()`

staticfloat:
This allows overloading things such as `ZMQ_DONTWAIT`.  Note that we explicitly mask out `ZMQ_SENDMORE` as it would make no sense to accept that in as one of the flags to `send()`.

### DIFF
--- a/zmq_addon.hpp
+++ b/zmq_addon.hpp
@@ -99,14 +99,14 @@ public:
         return m_parts.empty();
     }
 
-    bool recv(socket_t& socket)
+    bool recv(socket_t& socket, int flags = 0)
     {
         clear();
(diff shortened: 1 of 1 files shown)


bluca:
Hi,

Thanks for the pr, but isn't this an API breakage? Wouldn't it be preferable to follow the C4 process and have a new function instead?

c-rack:
Generally yes, you are right, but the new parameters have default values, so that existing implementations should not be affected. Please correct me if I'm wrong.

bluca:
Ah of course, it's a C++ api, my bad :-)

API will be fine, not sure about ABI though? I'll check

PR #83, opened 2016-08-26
Add comments, optimize some functions, add static asserts

K0n63n:
Add comments to all functions. Optimize move constructor and move assigment operator, remove size check inside pop/remove/peek. Static assert in pushtyp/addtyp/poptyp in order to avoid pitfall with std::string (pushstr/addstr/popstr should be used for that). Add sanity check for poptyp, size of message must match desired type.

### DIFF
--- a/zmq_addon.hpp
+++ b/zmq_addon.hpp
@@ -99,14 +99,14 @@ public:
         return m_parts.empty();
     }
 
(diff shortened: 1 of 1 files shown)


c-rack:
@K0n63n could you please resolve the conflict in order to make your PR mergable?

PR #85, opened 2016-09-19
Failing assertion in dtor of context_t

serge-medvedev:
Assertion in destructor of _context_t_ doesn't fail if _close_ method has already been called.


serge-medvedev:
This fix prevents assertion failures in case of calling _close_ method more than once and it seems to allow calling _close_ method after destruction, though it's not clear if such a behavior is desired one.
Also, it should fix [#74](https://github.com/zeromq/cppzmq/issues/74).

PR #86, opened 2016-10-22
Add comments, optimize some functions, add static asserts

K0n63n:
Add comments to all functions. Optimize move constructor and move assigment operator, remove size check inside pop/remove/peek. Static assert in pushtyp/addtyp/poptyp in order to avoid pitfall with std::string (pushstr/addstr/popstr should be used for that). Add sanity check for poptyp, size of message must match desired type.

### DIFF
--- a/zmq_addon.hpp
+++ b/zmq_addon.hpp
@@ -45,60 +45,72 @@ private:
     std::deque<message_t> m_parts;
 
 public:
+    // Default constructor
     multipart_t()
     {}
 
+    // Construct from socket receive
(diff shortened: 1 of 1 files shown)


bluca:
Thanks! For your next contributions, could you please follow the contributing guidelines for the git commit message? And also have one type of change of commit.

```
Problem: X is broken

Solution: do Y and Z to fix X
```

http://zeromq.org/docs:contributing

K0n63n:
Sure, I'll do that next time.

PR #107, opened 2017-01-09
Cmake files

JohanMabille:
This PR adds cmake files for installing cppzmq (and be able to find it via `find_package(cppzmq)`).

I've added version number in `zmq.hpp` and set it to `1.0.0` since there doesn't seem to be any tag / version number yet, and I don't know what the version policy should be.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+# Vim tmp files
+*.swp
+
+# Build directory
+build/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,15 +3,7 @@ project(cppzmq)
 
 find_package(ZeroMQ REQUIRED)
 
-file(STRINGS "${CMAKE_CURRENT_SOURCE_DIR}/zmq.hpp" cppzmq_version_defines
-    REGEX "#define CPPZMQ_VERSION_(MAJOR|MINOR|PATCH)")
-foreach(ver ${cppzmq_version_defines})
-    if(ver MATCHES "#define CPPZMQ_VERSION_(MAJOR|MINOR|PATCH) +([^ ]+)$")
-        set(CPPZMQ_VERSION_${CMAKE_MATCH_1} "${CMAKE_MATCH_2}" CACHE INTERNAL "")
-    endif()
-endforeach()
-set(${PROJECT_NAME}_VERSION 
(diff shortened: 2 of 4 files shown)


bluca:
Hi,

I think this header was intended to be used together with libzmq, so I'm not sure it should have its own build system with its own name and version

JohanMabille:
Just like any libzmq binding. However they all have their dedicated build, version number and packaging.

Actually this is not a full build system. The `nmake install` command only copies cppzmq headers to the right place and make it possible to find them with the `find_package` command.  And it is tightly coupled to libzmq since the libraries it exposes for linking are the ones of libzmq.

Without this, any package working with cppzmq would require a manual installation of it, making it impossible to have clean CI, scripted installation and so on.

SylvainCorlay:
> I think this header was intended to be used together with libzmq, so I'm not sure it should have its own build system with its own name and version

Hum, as soon as it is not distributed with the zmq package and can have its own updates, it makes sense to have independent versioning from zmq.

JohanMabille:
@bluca would it be ok if  I set the version to be the same as libzmq ? So we don't have to worry about it, and we still have the cmake packaging and installation feature.

bluca:
Yes please, that would make it clearer that they are intended to be used in lock steps

SylvainCorlay:
@bluca, I am looking at packaging cppzmq for conda.

Would you consider

 - tagging a release for cppzmq on GitHub
 - allowing cppzmq to have its own version numbering following semver (so that you can make minor releases fixing bugs independently of those of libzmq).

bluca:
First of all I am really not very involved with the development of this repository, basically I am a passer-by with strong opinions and access to the green buttons :-)

My main concern is compatibility. For example in CZMQ we have a build time and runtime detection system to ensure compatibility with all supported versions of libzmq. I don't think there's anything like that in this project.
Having the versions match means it's quite clear what works with what.

If you can confirm that the current version of cppzmq works with libzmq 4.1.2, I can tag with the same number.

SylvainCorlay:
OK I see where you come from. The required version range for libzmq could be specified in the cmake and the docs.  - It sounds like a good idea to start from the same number as libzmq itself.

Since libzmq follows semver, any major version number of libzmq should trigger a major version number change in cppzmq.

SylvainCorlay:
@bluca it does work - but we have mostly been using zeromq master. A github tag would be fantastic...

Issue #127, opened 2017-06-23
Missing FindZeroMQ.cmake


smeckels:
Attempting to build on Ubuntu 16.04 after installing ZeroMQ 4.2.1

> CMake Error at CMakeLists.txt:4 (find_package):
>   By not providing "FindZeroMQ.cmake" in CMAKE_MODULE_PATH this project has
>   asked CMake to find a package configuration file provided by "ZeroMQ", but
>   CMake did not find one.
> 
>   Could not find a package configuration file provided by "ZeroMQ" with any
>   of the following names:
> 
>     ZeroMQConfig.cmake
>     zeromq-config.cmake
> 
>   Add the installation prefix of "ZeroMQ" to CMAKE_PREFIX_PATH or set
>   "ZeroMQ_DIR" to a directory containing one of the above files.  If "ZeroMQ"
>   provides a separate development package or SDK, be sure it has been
>   installed.

> -- Configuring incomplete, errors occurred!

Am i invoking cmake incorrectly?

z3t0:
@smeckels Place [this](https://gist.github.com/z3t0/2c1c3fad2aeb65f10146fbc74a758d8a) somewhere in CMAKE_MODULE_PATH

z3t0:
Or [this](https://raw.githubusercontent.com/Kitware/Remus/master/CMake/FindZeroMQ.cmake) actually seems much more complete. 

Though this requires you to comment lines 6-7 in CMakeLists.txt for this library

stevemk14ebr:
Thank you. I've submitted a PR based on your solution, it does work well.

z3t0:
Just to be clear though this is what worked for me and I am not sure if commenting those 3 lines will have any other adverse side effects.

ax3l:
this should be fixed in the 4.2.2 release as of #133

Zvax:
I just got the same error from both downloading the 4.2.2 tar.gz file from the releases and checking out master. Being fairly new with cmake I'm not sure what I should include, but [here](https://gist.github.com/Zvax/f076f961fa9b27599aa198e69096b3e3) is the CMakeOutput.log and the error (which seems to be exactly the same as OP).

> CMake Warning at CMakeLists.txt:4 (find_package):
>   By not providing "FindZeroMQ.cmake" in CMAKE_MODULE_PATH this project has
>   asked CMake to find a package configuration file provided by "ZeroMQ", but
>   CMake did not find one.
> 
[...]

Issue #110, opened 2017-01-29
FindZeroMQ.cmake


abbradar:
Currently (in 4.2.1) cppzmq requires ZeroMQ module in CMake -- however, needed module is not bundled nor it can be found in libzmq distribution. Where should one take it from? Sorry if I have just overlooked something.

bluca:
It should no longer be necessary but I'm not a CMake expert. @JohanMabille any idea?

JohanMabille:
If you install cppzmq via cmake, you need to do the same with [libzmq](https://github.com/zeromq/libzmq)

Here is the command line for a minimal build and installation of libzmq:

```sh
cmake -D WITH_PERF_TOOL=OFF -D ZMQ_BUILD_TESTS=OFF -D ENABLE_CPACK=OFF
-D CMAKE_BUILD_TYPE=Release
make
make install
```

abbradar:
I did this but cmake modules were not installed. A quick search haven't shown anything similar in libzmq source distribution either. Where are they located? 
-- 
Nikolay.

JohanMabille:
They should be generated by the `make install` and the ZeroMQConfig.cmake.in file. It seems this file is available on master but not v4.2.1. Do you have the same issue with the master branch of libzmq ?

qzhang41:
Sir can you tell me where i go wrong? I build cppzmq and libzmq with cmake but it show the following erro:
CMake Error at D:/working this semester/libzmq-master1/ZeroMQConfig.cmake:19 (message):
  File or directory D://include referenced by variable ZeroMQ_INCLUDE_DIR
  does not exist !
Call Stack (most recent call first):
  D:/working this semester/libzmq-master1/ZeroMQConfig.cmake:36 (set_and_check)
  CMakeLists.txt:4 (find_package)

SylvainCorlay:
@abbradar did you specify the `CMAKE_INSTALL_PREFIX` when installing libzmq and zeromq.

abbradar:
Sorry for the late response. @JohanMabille is right -- file isn't present in ZeroMQ 4.2.1 but it's there in master (so master works for me).

PR #133, opened 2017-07-14
CMake: pkg-config Fallback

### LINKED ISSUE #132: Support libzmq's Autotools Install: pkg-config
Following #127 and https://github.com/zeromq/libzmq/issues/2621#issuecomment-314782300, the `libzmq` install *via autotools* is still a thing and used in [packaging systems](https://github.com/LLNL/spack/pull/4737), too.

Albeit the autotools install lacks the install of CMake config modules, it would be great if we could add support for `libzmq` autotools-installs anyway.

One way could be to support the always-installed `pkg-config` module named `libzmq` in `cppzmq`'s `CMakeLists.txt`.

For autotools based installs we could as a fallback to the current logic [piggyback on `pkg-config`](https://cmake.org/Wiki/CMake:How_To_Find_Libraries). This could look like this:
```cmake
find_package(PkgConfig)
pkg_check_modules(PC_LIBZMQ QUIET libzmq)

set(ZeroMQ_VERSION ${PC_LIBZMQ_VERSION})
find_library(ZeroMQ_LIBRARY NAMES libzmq.so
             PATHS ${PC_LIBZMQ_LIBDIR} ${PC_LIBZMQ_LIBRARY_DIRS})
find_library(ZeroMQ_STATIC_LIBRARY NAMES libzmq.a
             PATHS ${PC_LIBZMQ_LIBDIR} ${PC_LIBZMQ_LIBRARY_DIRS})
```

ax3l:
Try to find installs of `libzmq` that were performed with autotools with a `pkg-config` fallback. Close #132

@herbrechtsmeier we are trying to add a `pkg-config` fallback for `cppzmq`'s `CMakeLists.txt` to find `libzmq` installs that were done with autotools and thus have no ZeroMQ cmake module around.

Do you have an idea how we can re-introduce the targets you are using for the libs when found via `pkg-config`?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,35 @@
 cmake_minimum_required(VERSION 2.8.12)
 project(cppzmq)
 
-find_package(ZeroMQ REQUIRED)
+find_package(ZeroMQ)
+
+# libzmq autotools install: fallback to pkg-config
+if(NOT ZeroMQ_FOUND)
+    find_package(PkgConfig)
+    pkg_check_modules(PC_LIBZMQ QUIET libzmq)
+
+    set(ZeroMQ_VERSION ${PC_LIBZMQ_VERSION})
+    find_library(ZeroMQ_LIBRARY NAMES libzmq.so
+                 PATHS ${PC_LIBZMQ_LIBDIR} ${PC_LIBZMQ_LIBRARY_DIRS})
+    find_library(ZeroMQ_STATIC_LIBRARY NAMES libzmq.a
+                 PATHS ${PC_LIBZMQ_LIBDIR} ${PC_LIBZMQ_LIBRARY_DIRS})
+
+    add_library(libzmq SHARED IMPORTED)
+    set_property(TARGET libzmq PROPERTY INTERFACE_INCLUDE_DIRECTORIES ${PC_LIBZMQ_INCLUDE_DIRS})
+    set_property(TARGET libzmq PROPERTY IMPORTED_LOCATION ${ZeroMQ_LIBRARY})
+
+    add_library(libzmq-static STATIC IMPORTED ${PC_LIBZMQ_INCLUDE_DIRS})
+    set_property(TARGET libzmq-static PROPERTY INTERFACE_INCLUDE_DIRECTORIES ${PC_LIBZMQ_INCLUDE_DIRS})
+    set_property(TARGET libzmq-static PROPERTY IMPORTED_LOCATION ${ZeroMQ_STATIC_LIBRARY})
+
+    if(ZeroMQ_LIBRARY AND ZeroMQ_STATIC_LIBRARY)
+        set(ZeroMQ_FOUND ON)
+    endif()
+endif()
+
+if(NOT ZeroMQ_FOUND)
+    message(FATAL_ERROR "ZeroMQ was NOT found!")
+endif()
 
 if (ZeroMQ_FOUND AND (NOT TARGET libzmq OR NOT TARGET libzmq-static))
   message(FATAL_ERROR "ZeroMQ version not supported!")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.0.0)
 project(cppzmq)
 
 find_package(ZeroMQ)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,26 +5,7 @@ find_package(ZeroMQ)
 
 # libzmq autotools install: fallback to pkg-config
 if(NOT ZeroMQ_FOUND)
-    find_package(PkgConfig)
-    pkg_check_modules(PC_LIBZMQ QUIET libzmq)
-
-    set(ZeroMQ_VERSION ${PC_LIBZMQ_VERSION})
-    find_library(ZeroMQ_LIBRARY NAMES libzmq.so
-                 PATHS ${PC_LIBZMQ_LIBDIR} ${PC_LIBZMQ_LIBRARY_DIRS})
-    find_library(ZeroMQ_STATIC_LIBRARY NAMES libzmq.a
-                 PATHS ${PC_LIBZMQ_LIBDIR} ${PC_LIBZMQ_LIBRARY_DIRS})
-
-    add_library(libzmq SHARED IMPORTED)
-    set_property(TARGET libzmq PROPERTY INTERFACE_INCLUDE_DIRECTORIES ${PC_LIBZMQ_INCLUDE_DIRS})
-    set_property(TARGET libzmq PROPERTY IMPORTED_LOCATION ${ZeroMQ_LIBRARY})
-
-    add_library(libzmq-static STATIC IMPORTED ${PC_LIBZMQ_INCLUDE_DIRS})
-    set_property(TARGET libzmq-static PROPERTY INTERFACE_INCLUDE_DIRECTORIES ${PC_LIBZMQ_INCLUDE_DIRS})
-    set_property(TARGET libzmq-static PROPERTY IMPORTED_LOCATION ${ZeroMQ_STATIC_LIBRARY})
-
-    if(ZeroMQ_LIBRARY AND ZeroMQ_STATIC_LIBRARY)
-        set(ZeroMQ_FOUND ON)
-    endif()
+    include(${CMAKE_CURRENT_LIST_DIR}/libzmqPkgConfigFallback.cmake)
 endif()
 
 if(NOT ZeroMQ_FOUND)
@@ -79,4 +60,5 @@ install(EXPORT ${PROJECT_NAME}-targets
         DESTINATION ${CPPZMQ_CMAKECONFIG_INSTALL_DIR})
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}Config.cmake
               ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}ConfigVersion.cmake
+              ${CMAKE_SOURCE_DIR}/libzmqPkgConfigFallback.cmake
               DESTINATION ${CPPZMQ_CMAKECONFIG_INSTALL_DIR})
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -1,7 +1,24 @@
...
 
-find_package(ZeroMQ REQUIRED)
+find_package(ZeroMQ)
+
+# libzmq autotools install: fallback to pkg-config
+if(NOT ZeroMQ_FOUND)
+    find_package(PkgConfig)
+    pkg_check_modules(PC_LIBZMQ QUIET libzmq)
+
+    set(ZeroMQ_FOUND ON)
```
ax3l: does that need `CACHE` or even `FORCE`?

CMakeLists.txt
```diff
@@ -1,7 +1,24 @@
...
+                 PATHS ${PC_LIBZMQ_LIBDIR} ${PC_LIBZMQ_LIBRARY_DIRS})
+    find_library(ZeroMQ_STATIC_LIBRARY NAMES libzmq.a
+                 PATHS ${PC_LIBZMQ_LIBDIR} ${PC_LIBZMQ_LIBRARY_DIRS})
+endif()
+
+if(NOT ZeroMQ_FOUND)
+    message(FATAL_ERROR "ZeroMQ was NOT found!")
+endif()
 
 if (ZeroMQ_FOUND AND (NOT TARGET libzmq OR NOT TARGET libzmq-static))
```
ax3l: @herbrechtsmeier do you have an idea how I can define those targets again if `ZeroMQ_LIBRARY` and `ZeroMQ_STATIC_LIBRARY` are found (see above)

herbrechtsmeier:
I'm unsure if pkg-config is a good solution. But you could use the variables instead of the targets. But you need to call find_path to detect the include path.

I would recommand a FindZeroMQ.cmake and append its path to the CMAKE_MODULE_PATH. This could be installed by this project and could be used from the CMakeLists.txt and the cppzmqConfig.cmake.in. Otherwise you have to add the pkg-config fallback to both files.

You could also use the add_library function together with the IMPORTED signature. Therefore you have to copy and adapt more or less the code from the generated target files of libzmq.

ax3l:
Thanks for the feedback!

The first option is unlikely to happen since it was already dismissed in #110 & #127.

> You could also use the add_library function together with the IMPORTED signature. Therefore you have to copy and adapt more or less the code from the generated target files of libzmq.

thanks, likely a
```cmake
 add_library(libzmq IMPORTED)
 set_property(TARGET foo PROPERTY IMPORTED_LOCATION ZeroMQ_LIBRARY)
 add_library(libzmq-static STATIC IMPORTED)
 set_property(TARGET foo PROPERTY IMPORTED_LOCATION ZeroMQ_STATIC_LIBRARY)
```
will already solve it.

CMakeLists.txt
ax3l: added

CMakeLists.txt
ax3l: RT test worked as is

ax3l:
@bluca it works! :)

bluca:
Great stuff, thanks!
@herbrechtsmeier is this solution fine for you?

herbrechtsmeier:
Are you sure that it is correct to use the variables without `${}` inside the set_property function?

You don't touch the package configuration cppzmqConfig.cmake.in. Have you test to compile an external package which use find_package(cppzmq)?

Additionally you need to add the header path to the targets.

ax3l:
Thanks again for the review!

> Are you sure that it is correct to use the variables without ${} inside the set_property function?

True, following [this example page](https://cmake.org/Wiki/CMake/Tutorials/Exporting_and_Importing_Targets) it should use `${}`. Updated it, nevertheless I did not see a difference in the install dir.

> You don't touch the package configuration cppzmqConfig.cmake.in.

I didn't think that's not necessary when looking into it, but checking the installed `cppzmqConfig.cmake` shows we need to add the same logic there, yes!

> Have you test to compile an external package which use find_package(cppzmq)?

I just tested `make install` so far. This installed the two header files and the cmake files.

I now tested also a dependent project with a [`main.cpp`](http://zguide.zeromq.org/cpp:hwserver) and a `CMakeLists.txt` as follows:
```cmake
# project
cmake_minimum_required(VERSION 3.0.0)
project(ZeroMQexample)

# dependencies
find_package(cppzmq REQUIRED)

# executable
add_executable(main)

target_link_libraries(main ${cppzmq_LIBRARIES})
```

which still fails while looking for a `ZeroMQConfig.cmake` since the installed `cppzmqConfig.cmake` is not yet using the same logic. How can we add it there?

> Additionally you need to add the header path to the targets.

Sure, can you show me how please? :)

herbrechtsmeier:
Instead of the duplicated code I would recommend to move the code into its own cmake file and install and include this file in both files.

Even better would be a FindZeroMQ.cmake module. This would reduce the changes inside the `CMakeLists.txt` and `cppzmqConfig.cmake.in` to a `list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR})` before the `find_package` / `find_dependency` call and an additional `${CMAKE_CURRENT_BINARY_DIR}/FindZeroMQ.cmake` line inside the `install(FILES ...)` function.

I just released that there is already a FindZeroMQ.cmake inside an other ZeroMQ project which works without pkg-config:
https://github.com/zeromq/azmq/blob/master/config/FindZeroMQ.cmake

bluca:
Isn't the point of a lot of recent cmake changes in libzmq to remove the need for dependencies to ship a findzeromq file?

herbrechtsmeier:
@bluca The FindZeroMQ file is only needed for the autotools build system.

bluca:
Rather than shipping additional files which look like are needed by cmake but are actually for autotools, isn't it better to just use pkg-config which autotools already ships?

ax3l:
I just removed the code duplication.

ax3l:
> Even better would be a FindZeroMQ.cmake module

One could also make this a module, but in such case rather a `FindZeroMQFallback.cmake` module since one would not want to overwrite the default behavior to look for an existing "real" libzmq cmake module on `find_package`.

I think the solution for now is a good bridge between the autotools and the CMake world, so dependent projects can build safely with CMake.

For the long run, it's not a good idea to support two build systems in `libzmq` itself because one will always be bridging back and forth if the installs are not fully identically.

herbrechtsmeier:
The FindZeroMQ.cmake is needed by cmake and not by autotools.

The problem is that libzmq support different build systems and the autotools build don't install any cmake support. Hence all follow-up project needs do deal with the different build systems. Best would be to remove the autotools support from libzmq or let autotools install a FindZeroMQ.cmake.

Because of its limitations it isn't recommended to use pkg-config inside a cmake project.

herbrechtsmeier:
@ax3l You are right. You need to call `find_package(ZeroMQ CONFIG)` to bypass the FindZeroMQ.cmake module.

bluca:
Sorry, autotools is not going anywhere. Also work was recently done to remove the findzeromq file, so that's not coming back either.

pkgconfig is universal and well supported, so I would suggest sticking with that as the fallback, when the cmake files are not found.

ax3l:
I think we can live with that, now that we have a marvelous bridge in place 🦄

ax3l:
@bluca @herbrechtsmeier are you ok with the current state of the PR for a merge? :)

herbrechtsmeier:
@bluca What is the reason to support two build systems with different outputs? Why the autotools build don't provide also a native CMake support?

With this patch you use different code to detected libzmq inside czmq and cppzmq. The first project use a find package and the second project bypass the find package function but do similar things. This could be avoid with a common cmake support inside libzmq.

What do you mean by pkg-config is universal and well supported? There is a reason why cmake avoid the usage of pkg-config inside its modules.

ax3l:
@herbrechtsmeier I understand your pain but discussing this goes off-topic for this PR. Is it ok for you to just open a proposal in the issues of https://github.com/zeromq/libzmq? :)

bluca:
- We have to support multiple build systems, not only 2 (for example visual studio, and soon meson hopefully) because no single one can fit in all use cases, and more simply some users have different preferences and we have to respect that. Not everyone likes CMake for example.
An autotools-generated distributable tarball is the preferred delivery method for distributions, for embedded devices and for bootstrapping new architectures, as only a compiler is needed and the autotools-generated configure scripts are self-contained.
The tarball can't provide generated CMake files because that would require adding CMake to the toolchain and running it.

- I will add the same fallback to zproject soon, so that CZMQ and others can use it too. Given it's just a fallback in case the main method fails, which uses native CMake functions.

- All distributions and *nix systems support pkg-config, even OSX. Even 10 old versions of Solaris. We have to support all of them, and CMake is often not available, especially the latests versions that are required here. Again a distributable tarball that ships a pkg-config file will just work.
And it's the preferred method for all Linux distributions.

herbrechtsmeier:
Isn't visual studio supported by CMake? How many distributions don't offer a CMake package? CMake even supports bare metal cross compilation. Additionally the packages are relocatable and support dependency tracking. And nobody needs to use CMake for its project even if this project use CMake.

But even if you support different build systems you should support a common build system interface for follow up projects. You should really add CMake support to autotools. Otherwise you should remove the CMake package generation and only support pkg-config. But even this leads to a cmake file if it is called `libzmqPkgConfigFallback.cmake` or `FindZereMQ.cmake`. Additionally if the "autotools-generated distributable tarball is the preferred delivery method for distributions" the fallback will be the main interface.

Autotools could provide generated CMake files without CMake because it only needs to replace some variables inside a ZeroMQConfig.cmake text file like it does for pkg-config.

This project requires a three year old CMake version. Attentionally the pkg-config file could also be generated by CMake. 

With your changes you bring the FindZeroMQ.cmake file back. You only rename the file and use an include instead of the find_package(ZeroMQ) function. Additionally you distribute the file over different projects.

I could open a new issue in libzmq but this makes only sense if there is interests in an autotools support for CMake.

ax3l:
@bluca just to stay on topic of the PR: can you ping me pls when a `v4.2.2` of `cppzmq` is released so I can add a package to [spack](https://github.com/LLNL/spack) which we are using to build/ship our [downstream project](https://github.com/ComputationalRadiationPhysics/graybat) that brought me here? :)

PR #161, opened 2017-10-09
Fix compile error for libzmq 4.2.1 and 4.2.2

FrankAdesys:
When defined ZMQ_BUILD_DRAFT_API=1, you receive an error "'on_event_handshake_succeeded' was not declared in this scope"... which is correct for versions 4.2.1 and 4.2.2.
Function was renamed from 'on_event_handshake_succeed', fixed this by updating the function name.

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -872,7 +872,7 @@ namespace zmq
                 on_event_handshake_failed(*event, address.c_str());
                 break;
             case ZMQ_EVENT_HANDSHAKE_SUCCEED:
-                on_event_handshake_succeeded(*event, address.c_str());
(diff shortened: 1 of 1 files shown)


bluca:
It was called _succeed in 4.2.1 and 4.2.2:

https://github.com/zeromq/libzmq/blob/v4.2.2/include/zmq.h#L566

It's only in 4.2.3 that it was renamed

sigiesec:
This means that cppzmq client code must also be aware of the libzmq version, i.e. you need to use the same #if's when you need to support multiple versions. 

cppzmq does a simple 1-to-1 mapping of the libzmq features, so I don't think this should be changed.

PR #162, opened 2017-10-09
Revert "Fix compile error for libzmq 4.2.1 and 4.2.2"

sigiesec:
Reverts zeromq/cppzmq#161

I don't think this change is a good idea. The set of events changed between 4.2.2 and 4.2.3(unreleased). One change involved changing the name of ZMQ_EVENT_HANDSHAKE_SUCCEED to ZMQ_EVENT_HANDSHAKE_SUCCEEDED. The method names in cppzmq on_* correspond to the libzmq events and their names 1-to-1. 

This change removes this 1-to-1 mapping, which is confusing.

cppzmq makes no attempt to map different (DRAFT) API versions of libzmq to a uniform API. While this might appear okay for this identifier, which only contains a typo, a client library of cppzmq must be aware of the exact version of libzmq that is used.

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -872,7 +872,7 @@ namespace zmq
                 on_event_handshake_failed(*event, address.c_str());
                 break;
             case ZMQ_EVENT_HANDSHAKE_SUCCEED:
-                on_event_handshake_succeed(*event, address.c_str());
+                on_event_handshake_succeeded(*event, address.c_str());
                 break;
 #endif
 #endif


FrankAdesys:
I understand and totally agree what your are saying. I think it is a good thing to keep the same mapping. However, in this case there is a explicit section in the code that refers to versions >= 4.2.1 but < 4.2.3. In this section the define is indeed ZMQ_EVENT_HANDSHAKE_SUCCEED, however the function called is "on_event_handshake_**succeeded**(*event, address.c_str());".
 
(lines 857:877)
``` c++
#if ZMQ_VERSION >= ZMQ_MAKE_VERSION(4, 2, 3)
            case ZMQ_EVENT_HANDSHAKE_FAILED_NO_DETAIL:
                on_event_handshake_failed_no_detail(*event, address.c_str());
                break;
            case ZMQ_EVENT_HANDSHAKE_FAILED_PROTOCOL:
                on_event_handshake_failed_protocol(*event, address.c_str());
                break;
            case ZMQ_EVENT_HANDSHAKE_FAILED_AUTH:
                on_event_handshake_failed_auth(*event, address.c_str());
                break;
            case ZMQ_EVENT_HANDSHAKE_SUCCEEDED:
                on_event_handshake_succeeded(*event, address.c_str());
                break;
#elif ZMQ_VERSION >= ZMQ_MAKE_VERSION(4, 2, 1)
            case ZMQ_EVENT_HANDSHAKE_FAILED:
                on_event_handshake_failed(*event, address.c_str());
                break;
            case ZMQ_EVENT_HANDSHAKE_SUCCEED:                             <==== SUCCEED
                on_event_handshake_succeeded(*event, address.c_str());    <==== succeeded
                break;
#endif
```

First this is in conflict with containing the 1-to-1 mapping. 
Second, and for me more important, this produces inconsistent code for LIBZMQ versions 4.2.1 and 4.2.2, because later there is again an explicit section for these versions with definitions for the on_* functions. In this section, the naming is "_succeed" instead of the called "_succeeded".

(lines 912:920)
``` c++
#if ZMQ_VERSION >= ZMQ_MAKE_VERSION(4, 2, 3)
        virtual void on_event_handshake_failed_no_detail(const zmq_event_t &event_, const char* addr_) { (void) event_; (void) addr_; }
        virtual void on_event_handshake_failed_protocol(const zmq_event_t &event_, const char* addr_) { (void) event_; (void) addr_; }
        virtual void on_event_handshake_failed_auth(const zmq_event_t &event_, const char* addr_) { (void) event_; (void) addr_; }
        virtual void on_event_handshake_succeeded(const zmq_event_t &event_, const char* addr_) { (void) event_; (void) addr_; }
#elif ZMQ_VERSION >= ZMQ_MAKE_VERSION(4, 2, 1)
        virtual void on_event_handshake_failed(const zmq_event_t &event_, const char* addr_) { (void) event_; (void) addr_; }
        virtual void on_event_handshake_succeed(const zmq_event_t &event_, const char* addr_) { (void) event_; (void) addr_; }
#endif	
```

In conclusion: 

I agree to have the 1-to-1 mapping, but the master branch contains explicit sections for specific versions. Parts of these sections follow the 1-to-1 mapping strategy, other parts don't. It doesn't match and gives compilations errors. So please reconsider the choice to revert this patch, or remove these version-specific sections altogether, which makes sense as well.

sigiesec:
@FrankAdesys Argh, I messed this up completely, of course you are right.

PR #192, opened 2018-03-29
Add basic test infrastructure

sigiesec:


ci_build.sh
```diff
@@ -7,9 +7,9 @@ install_zeromq() {
 
     mkdir libzmq
     cd libzmq
-    curl -L https://github.com/zeromq/libzmq/releases/download/v${ZMQ_VERSION}/zeromq-${ZMQ_VERSION}.tar.gz >zeromq.tar.gz
+    curl -L https://github.com/zeromq/libzmq/archive/v${ZMQ_VERSION}.tar.gz >zeromq.tar.gz
```
bluca: if we want to keep people using cmake with the distributable tarball, maybe it should use it here as it would be a good test case

PR #196, opened 2018-04-07
Problem: poller_t does not have great test coverage

kurdybacha:
@sigiesec's recent commits with gtests and basics tests for zmq
types (thanks for that!) makes it too convinient to not add some more tests.

* Adding some simple tests for poller_t
* Stripped whitespaces from poller implementation.


sigiesec:
@kurdybacha Great, thanks a lot for that :) Wanted to add tests for poller_t myself, but didn't have the time yet!

kurdybacha:
@sigiesec no problem, you did heavy lifting here :) thanks!

PR #200, opened 2018-04-08
Problem: poller's handler not aware of event type.

kurdybacha:
It was possible to register a handler for more than one event types but
impossible to distinguish which event is being handled.
Now events are passed to a handler.

Besides that some other changes:
* new test covering changes in the poller
* modified existing tests to cover changes in the poller
* defined handler_t in poller_t scope for more convinient use
  and simpler code
* helper loopback binder to be re-used in tests

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -1020,7 +1020,9 @@ namespace zmq
             return *this;
         }
 
-        void add (zmq::socket_t &socket, short events, std::function<void(void)> &handler)
+        using handler_t = std::function<void(short)>;
+
+        void add (zmq::socket_t &socket, short events, handler_t &handler)
         {
             if (0 == zmq_poller_add (poller_ptr, socket.ptr, handler ? &handler : NULL, events)) {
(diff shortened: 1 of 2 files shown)


zmq.hpp
```diff
@@ -1020,7 +1020,9 @@ namespace zmq
             return *this;
         }
 
-        void add (zmq::socket_t &socket, short events, std::function<void(void)> &handler)
+        using handler_t = std::function<void(short)>;
+
+        void add (zmq::socket_t &socket, short events, handler_t &handler)
```
sigiesec: This should be done in a backwards-compatible way, i.e. keep a function with the old signature that wraps a std:: function<void(void)> and passes it to the function with the new signature.

kurdybacha: thank you for review. this is draft so no need for that.

sigiesec: Right. Since it is related to a libzmq draft api, it can be regarded draft in cppzmq as well.

Issue #199, opened 2018-04-08
CMake build fails on Windows


michael-hart:
Both on MinGW and using MSVC (Visual Studio 2017), I have build failures from being unable to build googletest. In both cases, removing the tests folder and the line from the top-level CMakeLists.txt to include the tests directory has caused successful compilation. I have included the build output as a pastebin.

https://pastebin.com/2sq2jv7N

michael-hart:
Seems related to google/googletest#1373 

In this issue thread, it is suggested to use:
```
cmake .. -DCMAKE_CXX_FLAGS=/D_SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING
```
This causes INSTALL to build correctly, but unit_tests still fails. The error log is here:
https://pastebin.com/kXMTT8rg

kurdybacha:
@michael-hart for reporting this and logs, they are very useful.
I think problem is here:
`error C2079: 'endpoint' uses undefined class 'std::array<char,100>'
and is caused by my last commit.
I do not have Windows machine with VS right now to verify by I believe that if you add:
`#include<array>` on top of poller.cpp that should fix the issue.
I will verify tomorrow.

kurdybacha:
#200 that has been merged to trunk might address the issue. Pending verification on Windows with VS.

sigiesec:
Appveyor CI with various VS versions, and maybe also gcc on mingw and cygwin should be added to catch such issues earlier. I planned to do this, but did not find the time yet.

michael-hart:
I've pulled from master and tried again. The flag I wrote in an earlier comment was still required, and the build still unfortunately fails. I've included the log:

https://pastebin.com/eavZa900

kurdybacha:
@michael-hart, I have just setup Windows machine with VS 2017 and can reproduce the issue.
Two problems with gtest:
1. Deprecation warning causing errors. You have already found a way to disable warning with `-DCMAKE_CXX_FLAGS=//D_SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING` or just `-DCMAKE_CXX_FLAGS=//w`.
2. By default, new Visual Studio projects link the C runtimes dynamically but Google Test links them statically. Workaround for now would be to build cppzmq with `-Dgtest_force_shared_crt=ON`.

My full commands to build cppzmq with cmake on Windows were:

```
$ cd cppzmq
$ mkdir build
$ cd build
[...]

PR #204, opened 2018-04-16
Problem: Windows build broken because of multiple issues

kurdybacha:
Windows issues:
* missing includes files
  Solution: added missing <memory> and <unordered_map>
  Here <map> was replaced with <unordered_map> as there is no need for
  sorted map.
* googletest fails because deprecation warning causing errors.
  Solution: D_SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING option added.
* googletest fails because by default, new Visual Studio projects
  link the C runtimes dynamically but Google Test links them statically.
  Solution: gtest_force_shared_crt=ON option added.

Two last point should fix issue #199.

Besides adding appveyor.yml configuration to add Windows build to
CI in order to prevent accidental Windows build breakage on merge.
For now only Debug configuration as Release requires more time to figure
out.

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -73,6 +73,7 @@
     #include <chrono>
     #include <tuple>
     #include <functional>
+    #include <unordered_map>
 #endif
 
 //  Detect whether the compiler supports C++11 rvalue references.
@@ -1114,7 +1115,7 @@ namespace zmq
     private:
         void *poller_ptr;
         std::vector<zmq_poller_event_t> poller_events;
-        std::map<socket_t*, handler_t> handlers;
+        std::unordered_map<socket_t*, handler_t> handlers;
     };  // class poller_t
 #endif //  defined(ZMQ_BUILD_DRAFT_API) && defined(ZMQ_CPP11) && defined(ZMQ_HAVE_POLLER)
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,5 +30,6 @@ build:
     verbosity: normal
 
 test_script:
+    - cp Build-libzmq/bin/%configuration%/libzmq*.dll Build/bin/%configuration/
     - cd Build
     - ctest -V -C %configuration%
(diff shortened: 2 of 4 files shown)


bluca:
I've enabled appveyor let's see if it runs

bluca:
Building libzmq on appveyor is extremely slow for some reasons...
If possible, you might want to disable building the tests as that might speed it up - as opposed to autotools, it looks like with cmake the tests are always built with "make" rather than "make test/check"

bluca:
The test fails:

```
1: Test command: C:\projects\cppzmq\Build\bin\Debug\unit_tests.exe
1: Test timeout computed to be: 10000000
1/1 Test #1: unit .............................***Exception: Exit code 0xc0000135
  0.20 sec
```

kurdybacha:
@bluca build it is really slow indeed, don't know why though yet.
I will check in the evening and come back on this. Thanks.

bluca:
Note that the slow build is a problem in libzmq's repository as well, so it's nothing to do with the appveyor file you added. Try with ```-DBUILD_TESTS=OFF```

kurdybacha:
I disabled tests and perfs so build time is 2x faster now (~5min). Proper way would be to "cache" googletest and libzmq build between runs. Until then I am thinking to disable VS 2015.
Failing ctest is still pending investigation.

bluca:
If you could implement ccache it would be fantastic, please also port it to libzmq and zproject :-)
It takes more than an hour right now to get a full run of the libzmq appveyor tests...

kurdybacha:
@bluca Windows build now passes on my branch. Should be fine here in a bit...

zmq.hpp
```diff
@@ -1107,7 +1108,7 @@ namespace zmq
     private:
         void *poller_ptr;
         std::vector<zmq_poller_event_t> poller_events;
-        std::map<socket_t*, handler_t> handlers;
+        std::unordered_map<socket_t*, handler_t> handlers;
```
sigiesec: I don't see how this change is related to the Windows build support (maybe #include <map> was missing?).

PR #207, opened 2018-04-17
Remove unnecessary curly bracket

kleisauke:
Will otherwise write `set(PACKAGE_VERSION "4.3.0}")` to `cppzmqConfigVersion.cmake`.


c-rack:
Thanks! For your next contributions, it would be great if you could follow our Problem/Solution style, see our C4 process.

PR #208, opened 2018-04-18
Problem: poller_t invalid behavior on moved sockets

kurdybacha:
On adding invalid socket (e.g. after move) there was exception thrown
but leaving modified and unconsistent internal state.
Besides that there was no possibility to remove a socket that was moved
into.

Solutions: changed internal unordered_map "handlers" to operator on zmq internal pointers and rollback when zmq_poller_add fails.

Added two test cases covering the issues.

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -536,6 +536,11 @@ namespace zmq
         {
             return ptr;
         }
+
+        inline operator bool() const ZMQ_NOTHROW
+        {
+            return ptr != NULL;
+        }
     private:
 
         void *ptr;
@@ -1053,10 +1058,12 @@ namespace zmq
 
         void add (zmq::socket_t &socket, short events, handler_t handler)
         {
+            if (!socket)
+                throw error_t ();
             handler_t *handler_ptr = nullptr;
             /// \todo is it sensible to allow handler to be empty? doesn't this lead to an error when the event is signalled? (perhaps it should already lead to an error in zmq_poller_add then)
             if (handler) {
-                auto emplace_res = handlers.emplace(&socket, std::move(handler));
+                auto emplace_res = handlers.emplace (socket.ptr, std::move (handler));
                 handler_ptr = &emplace_res.first->second;
             }
             if (0 == zmq_poller_add (poller_ptr, socket.ptr, handler_ptr, events)) {
@@ -1069,10 +1076,7 @@ namespace zmq
         void remove (zmq::socket_t &socket)
         {
             if (0 == zmq_poller_remove (poller_ptr, socket.ptr)) {
-                auto it = handlers.find (&socket);
-                if (it != handlers.end ()) { /// \todo this may only be false if handler was empty on add
-                    handlers.erase (it);
-                }
+                handlers.erase (socket.ptr);
                 poller_events.pop_back ();
                 return;
             }
@@ -1108,7 +1112,7 @@ namespace zmq
     private:
         void *poller_ptr;
         std::vector<zmq_poller_event_t> poller_events;
-        std::unordered_map<socket_t*, handler_t> handlers;
+        std::unordered_map<void*, handler_t> handlers;
     };  // class poller_t
 #endif //  defined(ZMQ_BUILD_DRAFT_API) && defined(ZMQ_CPP11) && defined(ZMQ_HAVE_POLLER)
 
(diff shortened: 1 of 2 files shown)


zmq.hpp
```diff
@@ -1053,10 +1058,12 @@ namespace zmq
 
         void add (zmq::socket_t &socket, short events, handler_t handler)
         {
+            if (!socket)
+                throw error_t ();
```
sigiesec: errno must be set to EINVAL for consistency, or error_t must be created with an explicit error code of EINVAL. An appropriate ctor of error_t must be added.

kurdybacha: Thank you for review. I don't think that setting EINVAL in cpp wrapper is a good idea, we have exceptions already for carrying information and should not rely on errno but I get the point that someone might rely on errno after throw already. Updated a review to throw only if zmq call fails so passing handling of nullptr socket to zmq.

sigiesec: The ctor of error_t reads errno. It is so far only intended to be thrown when there was a libzmq error.
As it is now, the exception will contain a random errno.

kurdybacha: Have you seen updated pull request?

zmq.hpp
```diff
@@ -1058,18 +1058,17 @@ namespace zmq
...
-            if (handler) {
-                auto emplace_res = handlers.emplace (socket.ptr, std::move (handler));
-                handler_ptr = &emplace_res.first->second;
-            }
-            if (0 == zmq_poller_add (poller_ptr, socket.ptr, handler_ptr, events)) {
+            auto it = std::end (handlers);
+            auto inserted = false;
+            if (handler)
+                std::tie(it, inserted) = handlers.emplace (socket.ptr, std::move (handler));
+            if (0 == zmq_poller_add (poller_ptr, socket.ptr, inserted ? &(it->second) : nullptr, events)) {
```
sigiesec: I don't think this is a sensible behaviour, if inserted == false. This would mean that add had already been called for the same socket, which is not a valid use. There probably should be a test for this case.

zmq.hpp
```diff
@@ -1058,18 +1058,17 @@ namespace zmq
...
-            handler_t *handler_ptr = nullptr;
-            /// \todo is it sensible to allow handler to be empty? doesn't this lead to an error when the event is signalled? (perhaps it should already lead to an error in zmq_poller_add then)
-            if (handler) {
-                auto emplace_res = handlers.emplace (socket.ptr, std::move (handler));
-                handler_ptr = &emplace_res.first->second;
-            }
-            if (0 == zmq_poller_add (poller_ptr, socket.ptr, handler_ptr, events)) {
+            auto it = std::end (handlers);
+            auto inserted = false;
+            if (handler)
```
sigiesec: The comment was removed, but it is still applicable.

zmq.hpp
kurdybacha: I think it is a valid use case and you are right that that if inserted is false it means socket already added but it will behave properly: rollback would be triggered, only exception with errno specific to "already present" scenario.

zmq.hpp
kurdybacha: Sorry, should have explained that. This comment was only confusing as it is valid to pass nullptr user data to zmq_poller_add. libzmq does not care what user passes there, it only passively handles it back on wait. Similarly to actors in czmq

sigiesec: It is valid for libzmq/zmq_poller_add, but it is not sensible in the way cppzmq uses it. In wait, a no-op results if handler was nullptr here.

kurdybacha: if you are referring to if we should allow empty handler at all then this is a separate discussion.
You should not create //todo comments for that because that might indicate that something was agreed (where?) and waiting to be done. If you think if something can be improved why not create an issue open for discussion, or provide pull request..
In my opinion we should allow empty handlers, it is a valid use case where user is interested in wait signalling only by returning and no-op.

sigiesec: My original comment was indeed mistaken. When I wrote it, I thought that it would lead to some error later on, but this is not the case. So I am fine to remove it.

If you used a poller with a single socket and an empty handler, it might make sense. But it is somewhat misaligned with the design of poller_t, which manages and calls back into the handlers. I would still tend to disallow is, but I am also fine to leave it as it is. It would be good to have a test that uses it this way (i.e. not only for adding an empty handler, but also polling afterwards).

zmq.hpp
sigiesec: Ok, fine then :) Sorry I missed that a test for this already exists.

PR #210, opened 2018-04-19
Problem: tests are always building

kleisauke:
Add the ability to explicitly not build tests. Option defaults to ON to keep the current behavior, but give the user the option to disable building of tests.

This option is similair to [`ZMQ_BUILD_TESTS`](https://github.com/zeromq/libzmq/blob/54fd20afdf0a5390b3e1e6a08b1f84f69d4446e4/CMakeLists.txt#L1096-L1100), but just for cppzmq.


kurdybacha:
looks good too me, thanks!

PR #211, opened 2018-04-20
Problem: poller_t's deprecated add might throw std::bad_function_call

kurdybacha:
Issue is reproducible in deprecated add method with empty handler
followed by wait that kicks in (covered by provided unit test).

I would prefer we remove this method completely as maintaining something
that we consider `deprecated` is unnecessary in `draft` API.


sigiesec:
I am fining with removing the overload.

PR #219, opened 2018-05-03
Problem: poller can segfault when modified from registered handler.

kurdybacha:
It is possible that a user would like to add/remove sockets from
handlers. As handlers and poll items might be removed while not
being processed yet - we have a segfault situation.
Provided unit test `remove_from_handler` demonstrates the problem.

Solution: Modify internal poll item data structure only after processing
of events is finished.

Please note that events processing path performance remains the same when there are
no modification (add/remove) to the poller (no rebuild) - main use case.

As an effect of changes `size` method has been removed as it does not
represent any meaningful information anymore. There are active and pending
(waiting for rebuild) poll items so two different sizes. User can
easily track on their side number of registered sockets if original size
information is needed.

`wait` method returns number of processed sockets now. It might be
useful information to a user for no extra cost.

### DIFF
--- a/tests/poller.cpp
+++ b/tests/poller.cpp
@@ -358,8 +358,9 @@ TEST(poller, remove_from_handler)
         }));
     }
     // Clients send messages
-    for (auto & s : setup_list)
+    for (auto & s : setup_list) {
         ASSERT_NO_THROW(s.client.send("Hi"));
+    }
 
     // Wait for all servers to receive a message
     for (auto & s : setup_list) {
--- a/tests/poller.cpp
+++ b/tests/poller.cpp
@@ -9,7 +9,6 @@
 TEST(poller, create_destroy)
 {
     zmq::poller_t poller;
-    ASSERT_EQ(0u, poller.size ());
 }
 
 static_assert(!std::is_copy_constructible<zmq::poller_t>::value, "poller_t should not be copy-constructible");
@@ -20,9 +19,7 @@ TEST(poller, move_construct_empty)
     std::unique_ptr<zmq::poller_t> a{new zmq::poller_t};
     zmq::poller_t b = std::move(*a);
 
-    ASSERT_EQ(0u, a->size ());
     a.reset ();
-    ASSERT_EQ(0u, b.size ());
 }
 
 TEST(poller, move_assign_empty)
@@ -32,9 +29,7 @@ TEST(poller, move_assign_empty)
 
     b = std::move(*a);
 
-    ASSERT_EQ(0u, a->size ());
     a.reset ();
-    ASSERT_EQ(0u, b.size ());
 }
 
 TEST(poller, move_construct_non_empty)
@@ -46,9 +41,7 @@ TEST(poller, move_construct_non_empty)
     a->add(socket, ZMQ_POLLIN, [](short) {});
     zmq::poller_t b = std::move(*a);
 
-    ASSERT_EQ(0u, a->size ());
     a.reset ();
-    ASSERT_EQ(1u, b.size ());
 }
 
 TEST(poller, move_assign_non_empty)
@@ -62,9 +55,7 @@ TEST(poller, move_assign_non_empty)
 
     b = std::move(*a);
 
-    ASSERT_EQ(0u, a->size ());
     a.reset ();
-    ASSERT_EQ(1u, b.size ());
 }
 
 TEST(poller, add_handler)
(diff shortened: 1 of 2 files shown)


zmq.hpp
```diff
@@ -1082,36 +1082,46 @@ namespace zmq
...
 #else
             if (zmq_errno () == ETIMEDOUT)
 #endif
-                return false;
+                return 0;
 
             throw error_t ();
         }
 
-        size_t size ()
```
sigiesec: I don't get why removing this method is necessary. There are several other options: get the size before pending changes, get the size after pending changes, provide getters for both.

sigiesec:
I am now quite sure that it makes sense to split poller_t into two layers: one that only manages socket registrations, and a separate one on top of that that knows about callback handlers.

tests/poller.cpp
```diff
@@ -9,7 +9,6 @@
 TEST(poller, create_destroy)
 {
     zmq::poller_t poller;
-    ASSERT_EQ(0u, poller.size ());
```
sigiesec: Something should be asserted here.

tests/poller.cpp
```diff
@@ -20,9 +19,7 @@ TEST(poller, move_construct_empty)
     std::unique_ptr<zmq::poller_t> a{new zmq::poller_t};
     zmq::poller_t b = std::move(*a);
 
-    ASSERT_EQ(0u, a->size ());
     a.reset ();
-    ASSERT_EQ(0u, b.size ());
```
sigiesec: There must be some assertion here. Or maybe the move operations of poller_t should be deleted entirely.

tests/poller.cpp
```diff
@@ -46,9 +41,7 @@ TEST(poller, move_construct_non_empty)
     a->add(socket, ZMQ_POLLIN, [](short) {});
     zmq::poller_t b = std::move(*a);
 
-    ASSERT_EQ(0u, a->size ());
     a.reset ();
-    ASSERT_EQ(1u, b.size ());
 }
```
sigiesec: Something must be asserted here. If there are no query operations, wait should be called on both pollers to ensure that only b has an effect.

tests/poller.cpp
```diff
@@ -46,9 +41,7 @@ TEST(poller, move_construct_non_empty)
     a->add(socket, ZMQ_POLLIN, [](short) {});
     zmq::poller_t b = std::move(*a);
 
-    ASSERT_EQ(0u, a->size ());
     a.reset ();
-    ASSERT_EQ(1u, b.size ());
```
sigiesec: Something must be asserted here. If there are no query operations, wait should be called on both pollers to ensure that only b has an effect.

zmq.hpp
kurdybacha: what value does it bring to the user? Would it not be more confusing than useful?
poller is not really a collection, you can not iterate through it etc. and user needs to store info about registered socket already in order to successfully unregistered it later.

tests/poller.cpp
kurdybacha: It is perfectly valid that unit test does not assert and runs without a crash/exception but if maintainers decide otherwise I am fine with that.

zmq.hpp
sigiesec: Maybe the size is not a relevant information, but the information if the poller is empty or not is, since this affects how wait may be called the next time. Therefore, this information should take into account that pending changes are applied. (I am not sure how you implemented handling of pending changes, but I think that there can only be pending changes if empty is called during an active call of wait, right? Actually, such reentrant calls are not relevant in the use cases I had in mind previously).

Overall I would be fine with removing the size method if an empty method is added instead.

tests/poller.cpp
sigiesec: I completely agree in principle, but I think in this case there should be some assertion.

This test would accept a move constructor that is a no-op...

But maybe it is fine if proper assertion are added for the non-empty case.

But still, I am not sure if it makes sense to provide move operations at all. If one wants to move a poller, they could use a unique_ptr to a poller_t instead.

zmq.hpp
kurdybacha: ..but `empty` has the same problem as `size`. For example:
```
zmq::poller_t poller;
poller.add (socket, ZMQ_POLLIN, [&](short events) {
    poller.remove (socket);
    bool empty = poller.empty();
    // hmmm
});
```
 What would you expect `empty` to be here? You are in the poller's event processing loop but just has removed last socket. If you still insist on adding `empty` or `size` I would go for `true` in the example above as it reflects latest user's actions.

kurdybacha: just created a PR for that.

tests/poller.cpp
kurdybacha: Everything inside the poller is movable already so it make it easy to keep poller movable too.

PR #220, opened 2018-05-08
Fix message using empty version variable

zguangyu:
project(cppzmq VERSION ${DETECTED_CPPZMQ_VERSION}) will create a variable called cppzmq_VERSION instead of CPPZMQ_VERSION. This is the case under cmake v3.9.4 .


kurdybacha:
Looks good to me. Thanks @zguangyu for providing the PR.

PR #224, opened 2018-05-11
Problem: design goals and supported platforms are not explicitly stated

sigiesec:
Solution: add section on design goals and supported platforms to README, improve overall structure of README


sigiesec:
Well, I extrapolated them from what was implemented, what impression I had what the goals might have been originally, and the reasons why we chose to use cppzmq.

If someone disagrees, we now have a basis for discussion. I could send a mail to the zeromq mailing list, asking for a discussion explicitly. What do you think about that?

kurdybacha:
Email seems reasonable, we might get some valuable inputs. Thanks!

PR #225, opened 2018-05-11
Split poller_t into two layers

sigiesec:
New poller_t (in zmq.hpp) is a direct mapping of the libzmq zmq_poller_* API.
New active_poller_t (in zmq_addon.hpp) is the previous poller_t, which is implemented based on the new poller_t, and provides a higher-level abstraction that actively manages callback handlers.

The PR contains some fixes to make zmq_addon.hpp compile again. It was broken before.


zmq_addon.hpp
```diff
@@ -408,190 +408,115 @@ class multipart_t
...
-
 inline std::ostream& operator<<(std::ostream& os, const multipart_t& msg)
 {
     return os << msg.str();
 }
 
+#endif  // ZMQ_HAS_RVALUE_REFS
+
+#if defined(ZMQ_BUILD_DRAFT_API) && defined(ZMQ_CPP11) && defined(ZMQ_HAVE_POLLER)
+    class active_poller_t
```
kurdybacha: Much cleaner now after moving this to separate class. Can we change a name though, please? Something more zmq'ish and shorter? Maybe loop_t, or service_t? I have some pending additions - timers - and it would make sense to put it here. Maybe with can split this actually into separate file? zmq_monitor.hpp, zmq_service.hpp?

zmq_addon.hpp
```diff
@@ -408,190 +408,115 @@ class multipart_t
...
+
+        void modify (zmq::socket_t &socket, short events)
+        {
+            base_poller.modify (socket, events);
+        }
+
+        int wait (std::chrono::milliseconds timeout)
+        {
+            if (need_rebuild) {
+                poller_events.resize (handlers.size ());
```
kurdybacha: This is wrong, sorry. resize does not clear items already in the vector so you might pass items from previous run to wait. Depending on this kind of stuff is unreliable and asking for trouble.

zmq_addon.hpp
```diff
@@ -408,190 +408,115 @@ class multipart_t
...
+
+        active_poller_t(active_poller_t&& src) = default;
+        active_poller_t &operator=(active_poller_t&& src) = default;
+
+        using handler_t = std::function<void(short)>;
+
+        void add (zmq::socket_t &socket, short events, handler_t handler)
+        {
+            auto it = decltype (handlers)::iterator {};
+            auto inserted = bool {};
```
kurdybacha: depending on default initialization here makes it only harder to read.

zmq_addon.hpp
```diff
@@ -408,190 +408,115 @@ class multipart_t
...
+            std::tie(it, inserted) = handlers.emplace (static_cast<void*>(socket), std::make_shared<handler_t> (std::move (handler)));
+            try 
+            {
+                base_poller.add (socket, events, inserted && *(it->second) ? it->second.get() : nullptr);
+                need_rebuild |= inserted;
```
kurdybacha: depending on default initialization here makes it only harder to read.

zmq_addon.hpp
```diff
@@ -408,190 +408,115 @@ class multipart_t
...
+            }
+            catch (const zmq::error_t&) 
+            {
+                // rollback
+                if (inserted)
+                {
+                    handlers.erase (static_cast<void*>(socket));
```
kurdybacha: static_cast is unnecessary here.

zmq_addon.hpp
```diff
@@ -408,190 +408,115 @@ class multipart_t
...
+                poller_events.resize (handlers.size ());
+                poller_handlers.clear ();
+                poller_handlers.reserve (handlers.size ());
+                for (const auto &handler : handlers) {
+                    poller_handlers.push_back (handler.second);
+                }
+                need_rebuild = false;
+            }
+            const int count = base_poller.wait_all (poller_events, timeout);
+            if (count != 0) {
```
kurdybacha: count > 0

tests/active_poller.cpp
```diff
@@ -0,0 +1,414 @@
...
+    zmq::active_poller_t::handler_t handler = [&message_received](short events) {
+        ASSERT_TRUE(0 != (events & ZMQ_POLLIN));
+        message_received = true;
+    };
+    ASSERT_NO_THROW(active_poller.add(s.server, ZMQ_POLLIN, handler));
+    ASSERT_EQ(1, active_poller.wait(std::chrono::milliseconds{-1}));
+    ASSERT_TRUE(message_received);
+}
+
+/// \todo this contains multiple test cases that should be split up
```
kurdybacha: can we remove this comment?

zmq.hpp
```diff
@@ -1019,75 +1018,42 @@ namespace zmq
     };
 
 #if defined(ZMQ_BUILD_DRAFT_API) && defined(ZMQ_CPP11) && defined(ZMQ_HAVE_POLLER)
+    template <typename T = void>
```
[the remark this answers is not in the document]
sigiesec: It makes it typesafe when adding. active_poller_t uses its handler_t as the template argument. There is at least one test using int as an argument. I think it makes sense.

tests/poller.cpp
```diff
@@ -8,117 +7,98 @@
 
 TEST(poller, create_destroy)
 {
-    zmq::poller_t poller;
-    ASSERT_TRUE(poller.empty ());
+    zmq::poller_t<> poller;
```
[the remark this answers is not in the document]
sigiesec: I somehow waited for this question ;)

Well, poller_t has no state of its own, that could or would need to be tested. zmq_poller_* do not allow to query anything at the moment. Maybe that would be a good idea, and then an assertion using that could be added here.

zmq_addon.hpp
sigiesec: I am not completely satisfied with my name as well. I think "poller" should be part of the name, but "active" is not so specific. loop_t would be strange IMO, as there is no loop within wait. A "service" is also something quite different for me, which has its own thread etc. These concepts could be built on top of active_poller_t again. 

About splitting up the headers... I am somewhat undecided about that. It is a feature of cppzmq that it consists only of two header files. But on the other hand, when you can copy two header files, you can also copy a handful. If it is done, zmq_addon.hpp should include all the extra headers then.

zmq_addon.hpp
sigiesec: Hm. I don't think so. Initializing with a value explicitly for which there is no code path where it could ever be read would be confusing IMO.

zmq_addon.hpp
sigiesec: That's why I changed this. It is not necessary to clear the items. zmq_poller_wait_all even clears all unused items. But this is unnecessary too. zmq_poller_wait_all returns the number of valid items, and we only access those.

If this were different, you sould need to clear all items on every call to wait, regardless of the value of need_rebuild. But this would be quite inefficient if you had a large pollset.

zmq_addon.hpp
sigiesec: In order to avoid an unnecessary rebuild?
IIRC it was like this before, just something like
```
if (inserted)
{
  need_rebuild = true;
  ...
}
```

zmq_addon.hpp
sigiesec: Yes, I agree, that would be better.

tests/active_poller.cpp
sigiesec: Do you think it no longer applies?

zmq_addon.hpp
sigiesec: Actually, the condition is completely redundant. The for_each below just iterates over an empty range if count == 0.

zmq_addon.hpp
kurdybacha: yeah, `naming` is the most difficult part in software development. maybe zpoller_t (z - zuper) :)
We can stay with two header for now. If have more stuff we can split then and I like an idea of zmq_addon.hpp to have all , probably with a rename to zmq_addons.hpp

zmq_addon.hpp
kurdybacha: I get a point about performance but my issue was that we rely on internal implementation of libzmq to clear them (set all members) for us. How about if `libzmq` decides to set only polleritem partially depends on the case e.g. ZMQ_POLLERR etc.. That was my original reason behind clear().

zmq_addon.hpp
kurdybacha: There is not difference in initializing it explicitly or default. But there is readability and maintainability gain. Explicit initialization makes it harder to break things in the future IMO. Your call here.

zmq_addon.hpp
kurdybacha: You not avoiding here anything. If you are after `base_poller.add` you have to rebuild regardless of `inserted`.

tests/active_poller.cpp
kurdybacha: I think it has never applied. All separate cases are already covered? This is a test that exercises more complex scenario - splinting it up does not make sense because you loose the indent of it.

zmq_addon.hpp
sigiesec: Hm, I think add will throw necessarily if inserted is false, so the need_rebuild line will only be reached if inserted is true anyway. So... it could be changed...

zmq_addon.hpp
sigiesec: I can understand your point of view regarding future changes. I disagree about readability though. If I read some variable is initialized with a specific value I assume this value will be used in some case, and search for a long time if there is no such case.

Anyway, there is no way without some disadvantage. What is more important is to do this consistently throughout cppzmq IMO (I am not sure if there is any other comparable place at the moment). We should prescribe either way in a brief style guide.

tests/active_poller.cpp
sigiesec: Ok, then remove it.

zmq_addon.hpp
sigiesec: I think the previous version did not clear the members of the event struct either.

We shouldn't rely on implementation details of course, but I don't think we do. We should rely on the specification of the API, which should specify that the indicated number of entries are properly set to be well usable IMO. However, I fear the zmq_poller_* functions are not properly documented yet. I will put this forward on the libzmq side.

kurdybacha: Sounds good.

zmq_addon.hpp
kurdybacha: Ok. Agree.
We would not have this discussion if we switch poller to C++14. What do you think about that?

sigiesec: I don't think this would be a good idea at the moment. I read that people can't use fbzmq e.g. because it requires C++14 support. For our own uses we will stick to VS2015 for several years to come e.g., which has only very limited support for C++14 or even C++17 features. I think there are also many other environments that at least have no recent gcc/clang by default. We already restrict the platforms quite substantially compared to libzmq itself which does not require C++11.

PR #226, opened 2018-05-11
Problem: inconsistent code style

kurdybacha:
Solution: copy .clang-format from `libzmq` as a starting point.

There is no need to follow exactly what libzmq does as this is header only library but it seems reasonable starting point. It would be good to stick to one style as there is a lot of activity and good ideas lately.
Build changes to follow.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,53 @@
+BasedOnStyle: LLVM
+IndentWidth: 4
+UseTab: Never
+BreakBeforeBraces: Custom
+BraceWrapping: 
+    AfterClass: true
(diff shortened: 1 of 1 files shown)


kurdybacha:
please ignore coverage run with this changes as it is caused by line breaks. Column is now 80 as per clang-format.

kurdybacha:
I would like to run this with you. What do you think about two style changes:
`SpaceBeforeParens: Always` -> SpaceBeforeParens: false
`ColumnLimit: 80 -> ColumnLimit: 85` //as this is header only library and everyone has at least 21inch monitors now

PR #232, opened 2018-05-16
Problem: libzmq pkg-config build not covered by CI

kurdybacha:
Solution: libzmq pkg-config build (using autogen) added to Travis CI.

There is an issue with cppzmq where it needs to be installed as well in order to find pkg-config version of libzmq. That will be addressed in separate PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,14 +22,14 @@ matrix:
     # GCC on Linux
     ##########################################################
 
-    # GCC default, draft disabled, latest libzmq
-    - os: linux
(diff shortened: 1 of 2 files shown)


bluca:
You can install to a local directory and avoid the sudo

kurdybacha:
true, will change that. thanks.

kurdybacha:
@bluca, not that easy as I thought. for some reason cppzmq does not compile with locally installed pkg-config version of libzmq. I would need to spend more time on that some other day.

PR #234, opened 2018-05-18
Problem: cppzmq needs to be installed for pkg-config libzmq

kurdybacha:
Solution: Make sure that FinZeroMQ.cmake is present in cppzmq's build
(binary) directory. That way cppzmq can be used from build/binary directory directly for all the cases.


bluca:
Can the sudo be removed from the Travis config file now?

kurdybacha:
not yet, we still need to install libzmq in pkg-config case.

PR #235, opened 2018-05-19
Add message_t::routing_id() and set_routing_id()

### LINKED ISSUE #233: Draft API methods need to be exposed
I'd like to make use of ZMQ_CLIENT / ZMQ_SERVER socket types with cppzmq, but unfortunately I am not able to, as cppzmq doesn't expose zmq_msg_routing_id() / zmq_msg_set_routing_id(). I can't simply call those functions directly, as zmq::message_t doesn't expose the underlying zmq_msg_t type.

Tulon:
Setting a routing id is necessary when sending a message through a
ZMQ_SERVER socket. See [1] for more details.

This commit fixes #233.

[1] http://api.zeromq.org/4-2:zmq-socket#toc5

### DIFF
--- a/tests/message.cpp
+++ b/tests/message.cpp
@@ -118,3 +118,12 @@ TEST(message, equality_non_equal_lhs_empty)
     const zmq::message_t msg_b("Hi", 2);
     ASSERT_NE(msg_a, msg_b);
 }
+
+#if defined(ZMQ_BUILD_DRAFT_API) && ZMQ_VERSION >= ZMQ_MAKE_VERSION(4, 2, 0)
+TEST(message, routing_id_persists)
+{
+    zmq::message_t msg;
+    msg.set_routing_id(123);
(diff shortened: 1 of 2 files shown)


Tulon:
I am not quite sure why some CI builds fail. It looks like ZMQ_BUILD_DRAFT_API may be defined even though libzmq was built with draft APIs disabled.

kurdybacha:
@Tulon thank you for PR. I would be able to take a look in the evening (GMT). Maybe someone else can help in the meantime. I would suggest you try to use `ci_build.sh` locally. Just run `./ci_build.sh` without any arguments  as this is one of the failing use cases. Maybe you get more info.

sigiesec:
Hm, the poller stuff is additionally guarded by ZMQ_HAVE_POLLER, so maybe ZMQ_BUILD_DRAFT_API might really be always set. This might be an error on either the libzmq or our side.

PR #236, opened 2018-05-20
Problem: cppzmq build broken with C++11 compiler and git cloned repo.

kurdybacha:
Default build, ./ci_build.sh without any arguments, which causes to run
with draft disabled, does not work properly for git cloned repository and
C++11 compiler.

Two issues:
1. For git cloned repository ENABLE_DRAFTS is ON by default but libzmq
compiled build without drafts .Travis did not catch that because default build
runs on non C++11 compiler.
2. testutil.hpp does not build because of missing draft guards.

Solution 1: Remove check for presence of .git for enabling draft API as it
is confusing to use with ENABLE_DRAFTS flag and there should be only one
explicit way to enable draft build. That fixes #235 wher draft was implicitly enabled.

Solution 2: add missing draft guards in testutil.hpp for server/client
socket in use there.

Solution 3: add extra Travis build covering C++11 compiler and non
draft build.


CMakeLists.txt
```diff
@@ -19,11 +19,7 @@ if (ZeroMQ_FOUND AND (NOT TARGET libzmq OR NOT TARGET libzmq-static))
   message(FATAL_ERROR "ZeroMQ version not supported!")
 endif()
 
-if (EXISTS "${CMAKE_SOURCE_DIR}/.git")
```
bluca: Note that this check is used by all other ZMQ projects which implement a draft system

kurdybacha: I see, in this case we can keep it consistent with others.

PR #238, opened 2018-05-21
feat: disable tests by default if not main project

gocarlos:
my disabling the unit tests by default if compiling this project as part of another project using e.g. add_subdirectory() we don't need internet for google test which is currently added as an external project. 

if the project is compiled as a stand alone project, the unit tests will be compiled by default.


gocarlos:
Hi Pawel, 
my bad, I tried doing: 
```
set(CPPZMQ_BUILD_TESTS OFF)
add_subdirectory(${CMAKE_CURRENT_LIST_DIR}/third_party/cppzmq)
```
which did not work.... 

the correct way is 
```
set(CPPZMQ_BUILD_TESTS OFF CACHE BOOL "Whether or not to build the tests")
add_subdirectory(${CMAKE_CURRENT_LIST_DIR}/third_party/cppzmq)
```

so no change is needed here... 
closing PR...

kurdybacha:
Great. btw I would suggest you use ExternalProject_Add for you third party dependencies.

PR #246, opened 2018-06-02
Problem: coverate not supported

kurdybacha:
Solution: enable code coverage and report it to codecov.io
Note: This requires zeromq/cppzmq to be enabled on codecov.io

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,21 @@ matrix:
             - g++-7
       env:
         - MATRIX_EVAL="CC=gcc-7 && CXX=g++-7" ENABLE_DRAFTS=ON
+
+    # Coverage, GCC 7, draft enabled, latest libzmq (default)
+    - os: linux
+      addons:
+        apt:
+          sources:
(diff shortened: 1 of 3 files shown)


bluca:
Why the change from coveralls?

kurdybacha:
just found it easier to integrate and easier to work with (subjective). If that is an issue I can go back to coverall, no problem.

bluca:
Just curious on the pro and cons,I don't really mind or way or the other

kurdybacha:
I will make coveralls work on a separate branch and we can compare and make more informative decision

kurdybacha:
@bluca just created another PR with `coveralls`. Both `coveralls` and `codecov` show same results now.
With codecov there is some more work required as it should not show `tests`.
Overall both are good. I don't have strong opinion so whatever you merges is fine with me.
I found codecov integration easier (no need for pip install, simpler usage) and UI seems more readable (e.g. similar fonts as on github) but as `libzmq` uses `coveralls` maybe we can keep it consistent.

PR #247, opened 2018-06-03
Problem: Coverage with coveralls not working

kurdybacha:
Solution: Enable coveralls for one linux build target and fix coveralls
configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,21 @@ matrix:
     # GCC on Linux
     ##########################################################
 
+    # Coverage, GCC 7, draft enabled, latest libzmq (default)
+    - os: linux
+      before_install:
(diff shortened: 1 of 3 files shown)


sigiesec:
@kurdybacha Thanks a lot for working on this. Unfortunately, it is still not right. For example, https://coveralls.io/builds/17296150/source?filename=zmq.hpp#L737 is still not shown as uncovered. Indeed, I think (almost?) all functions that are not used do not appear in the coverage report. I tried a bit around with lcov to get this working, following the hints at https://stackoverflow.com/a/19792966, but it did not work out either :(

kurdybacha:
Yes, it is still broken :( will take a look again on this. Thanks for the link, it might be helpful.

PR #256, opened 2018-07-14
Add support for RADIO/DISH sockets if draft API is enabled

Tulon:
This commit introduces new socket_type enumeration values as well
as the following supporting functions:

socket_t::join()
socket_t::leave()
message_t::group()
message_t::set_group()


sigiesec:
Thanks!

PR #264, opened 2018-08-15
Update version to 4.4.0 for next development iteration

sigiesec:


bluca:
If there are no new APIs been added, it can be set to 4.3.1 for now

sigiesec:
I think it is more likely that there will be changes that go beyond bugfixes, so 4.4.0 is better IMO.

PR #265, opened 2018-08-21
Problem: addition of new single-argument message_t introduces ambigui…

sigiesec:
…ty when calling e.g. socket_t::send

Solution: make single-argument constructor explicit

### DIFF
--- a/tests/active_poller.cpp
+++ b/tests/active_poller.cpp
@@ -157,7 +157,7 @@ TEST(active_poller, poll_basic)
 {
     server_client_setup s;
 
-    ASSERT_NO_THROW(s.client.send("Hi"));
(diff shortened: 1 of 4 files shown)


sigiesec:
@bluca can you give this a short look and merge it if you see no issues?

kurdybacha:
Hi Simon,

thanks for looking at this. Could you explain ambiguity that you experience?
I am not sure if `explicit` specifier with templates arguments helps here..

sigiesec:
It does. I tested it with our old code. You can have a look at the new test case that I added, which is something that worked before. When you remove the explicit again, you will get an ambiguous function call error.

kurdybacha:
Got it. `inline bool recv(message_t *msg_, int flags_ = 0)` is called in this case. Thanks.

PR #268, opened 2018-09-23
Problem: Dependency on googletest framework

kurdybacha:
Currently cppzmq as relatively simple and header only library depends on rather
complex unit test framework googletest.
Current issues:
- Googletest requires downloading and building it every time on travis
as cache support is limited there
- Googletest build is signifficant with comparison to cppzmq unittests
total runtime

Solution: Port existing tests to Catch - header only C++ framework and
gain ~20% build speed up on travis.

Why Catch?
It is well know C++ header only testing framework. It works well, it is
being kept up to date and maintainers seem to pay attention to
community's comments and issues.
We can not use Catch2 currently as we still support pre-C++11 compilers.


kurdybacha:
Hi @sigiesec , thanks for looking at this.

Both Catch and Doctest have the same philosophy:

>Most test frameworks have a large collection of assertion macros to capture all possible conditional forms (_EQUALS, _NOTEQUALS, _GREATER_THAN etc). doctest is different (but it's like Catch in this regard). Because it decomposes comparison expressions most of these forms are reduced to one or two that you will use all the time.

Besides in the thread `[zeromq-dev] cppzmq revival and RFC on design goals and supported platform` you said:

>Personally I have no particular preference for googletest, and at first sight I think we could go with doctest as well. I agree that not needing a build step is a plus, in particular for a project like cppzmq, which is a header-only library itself. However, someone would need to migrate the existing tests. Would you be willing to do that? That would be great.

So why change of opinion now? Only because of equality assertion?
For C/C++ programmers, and cppzmq is C/C++ project, I think it is clear and readable to have  `CHECK(0u == message.size())` and I and people I have worked with have not had issue with that.
And googletest `ASSERT_EQ` just does the same: `ASSERT_EQ(val1, val2) => val1 == val2`

kurdybacha:
I forgot to mention. Catch give you this:
```
message.cpp:15: FAILED:
  CHECK( 1u == message.size() )
with expansion:
  1 == 0
```
so you get the actual values.

kurdybacha:
@sigiesec ? any opinion @bluca @c-rack ?

bluca:
I don't really use cppzmq so I leave it to you folks :-)

kurdybacha:
@sigiesec Could you address comments above, please?

sigiesec:
@kurdybacha I am very sorry for taking so long to answer, I was too busy in other contexts unfortunately. I really think that the output for equality assertions is an important point. It allows you to avoid starting a debugging session for many test failures. I spent too much time in my life doing this. But as you pointed out, the actual values are output by CATCH, so everything is fine :)

sigiesec:
Does this need to be updated to accomodate for the latest changes on the master?

kurdybacha:
Hi @sigiesec . Thanks. Yes, just rebased it against latest master.

QuLogic:
`ctest -V` output seems a lot emptier now compared to 4.3.0.

kurdybacha:
thanks @QuLogic, looks like something that we can improve.

kurdybacha:
have a fix for that, will create PR once current pending one merges. I wish github had dependent PR support...

PR #269, opened 2018-10-16
add front and back

tangfu:
i'll get the last message in some situations，but have to do like this :
```
multimsg msgs;
auto msg = msgs[msgs.size() - 1];
func(msg.size());
auto *msg = msgs.end() - 1;
```
but, std::queue have some method 'front' and 'back'. so maybe i can simply do this ？ :
```
multimsg msgs;
auto msg = msgs.back();
```


kurdybacha:
Thanks for that. If you would find some more time it would be great to add some tests for these new methods (`tests/multipart.cpp`)

Issue #171, opened 2017-12-01
zmq::context_t dtor does not handle EINTR


andrust:
http://api.zeromq.org/master:zmq-ctx-destroy states:
"Return value

The zmq_ctx_destroy() function shall return zero if successful. Otherwise it shall return -1 and set errno to one of the values defined below.

Errors

EFAULT
The provided context was invalid.
EINTR
Termination was interrupted by a signal. It can be restarted if needed."

sigiesec:
Could you provide an example usage, and state what behaviour you expect? Should the dtor retry in that case?
You are also welcome to send a PR which adds a test case, and implements the expected behaviour.

egpbos:
I am mainly interested in handling EINTR, which I'm currently running into (as mentioned in #28). I will think about how to trigger this signal, but not sure whether I can manage, because for me it is currently a bug, so I don't fully understand it yet.

andrust:
I don't remember the exact situation how i stumbled upon this issue, but i think a signal was interrupting the close function during shutdown.

https://github.com/zeromq/cppzmq/blob/c444c5d5e2ecd9fef2da2455809599bfa34347fd/zmq.hpp#L518-L528
```cpp
    inline ~context_t() ZMQ_NOTHROW { close(); }

    inline void close() ZMQ_NOTHROW
    {
        if (ptr == NULL)
            return;

        int rc = zmq_ctx_destroy(ptr);
        ZMQ_ASSERT(rc == 0);
        ptr = NULL;
    }
```

In above snippet, i think we should retry `zmq_ctx_destroy` when `errno == EINTR` as the documentation states. It is unclear to me what the `if needed` part means. Maybe someone, who's familiar with the C code could make it clear.

egpbos:
I also thought just retrying is probably the right solution, but just to be sure, I dove into the C code, and now I'm not so sure anymore.

[...]

PR #271, opened 2018-11-01
WIP: adding test that fails on context.close() with EINTR

egpbos:
This is a work in progress PR to address issue #171.

I tried to make a minimal reproduction of the conditions under which my own code is currently failing. It fails seemingly randomly with an EINTR on closing the context.

The test I added in this PR fails for me *only* when run from the `lldb` debugger, compiled on macOS 10.14 with XCode clang 10.0.0 (clang-1000.11.45.2). It then indeed reports that it failed because of an interrupted system call (= EINTR).

I was thinking it would be ideal if we can manually trigger such an interrupt on the relevant system call made from zmq_ctx_destroy. However, I have no idea how to do this. Also, I don't know exactly which system call it is that is sometimes being interrupted.

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -102,7 +102,7 @@
 
 /*  Version macros for compile-time API version detection                     */
 #define CPPZMQ_VERSION_MAJOR 4
-#define CPPZMQ_VERSION_MINOR 6
+#define CPPZMQ_VERSION_MINOR 7
 #define CPPZMQ_VERSION_PATCH 0
 
 #define CPPZMQ_VERSION                                                              \


sigiesec:
@egpbos I appreciate your rigorous approach to the problem, but I fear that there is no portable way to trigger the EINTR error (and I don't know a non-portable way for Linux or Mac OS X either). The test case won't compile under Windows and other OSs that do not have unistd.h at all.

I think it should be fine if the dtor is fixed by adding a loop that continues while the call fails with EINTR, review the change thoroughly, and do not explicitly have a test for it. If you can do this, this would be great.

I think the only way to test this portably would be to use some instrumented libzmq, which allows to trigger the situation, but I think that would be too complex for the benefits.

egpbos:
You may be right! Just to be sure, I posted [a question on Stack Overflow](https://stackoverflow.com/questions/53119771/can-a-system-call-be-programmatically-interrupted). Let's see if it gets some interesting comments. If not, your suggestion sounds good; I will try to implement a retry.

egpbos:
The Stack Overflow discussion did not yet give the solution I was looking for. A colleague suggested mocking either the system call (poll or select in this case, see [analysis here](https://github.com/zeromq/cppzmq/issues/171#issuecomment-435943945)) or `zmq_ctx_destroy`, manually setting `EINTR` in the mocked version. This could be a possible way forward for this, but it may still be a little too much trouble, given the discussion in #171. What do you think @sigiesec?

egpbos:
I tried it, but it doesn't work. The problem is timing: you have to produce the SIGALRM at the exact time that the system call (poll or select) is being run. The random delays caused by other system processes make this impossible: most SIGALRMs will be triggered before or after the system calls.

egpbos:
The approach I think is most promising is the `ptrace` one I linked to on the SO page in the comments. It only works on Linux though, and I'm developing on macOS. I think that would be quite a nice general technique, indeed, for this kind of debugging.

Issue #280, opened 2018-11-26
during make install: catch.hpp alternative option


tenfoldpaper:
Hello,

I am currently trying to install cppzmq on a company computer with blocked access to github domain from terminal (I can access it fine from a browser, however). As such, invoking make -j4 install fails due to the script trying to download the catch.hpp file from the github domain. Downloading the file manually and putting it in the correct location doesn't fix anything since the URL_HASH check still fails. 

Any help would be greatly appreciated.

-- File already exists but no hash specified (use URL_HASH):
  file='/home/a27506446/Downloads/cppzmq-master/build/catch/src/catch.hpp'
Old file will be removed and new file downloaded from URL.
-- Downloading...
   dst='/home/a27506446/Downloads/cppzmq-master/build/catch/src/catch.hpp'
   timeout='none'
-- Using src='https://raw.githubusercontent.com/catchorg/Catch2/Catch1.x/single_include/catch.hpp'
-- Retrying...
-- Using src='https://raw.githubusercontent.com/catchorg/Catch2/Catch1.x/single_include/catch.hpp'
-- Retry after 5 seconds (attempt #2) ...

PR #288, opened 2019-01-26
Problem: catch must be downloaded during build

QuLogic:
This is trouble for building packages as there is no network, and we want everything to be available from sources we've already built.

The solution here is to allow building against an external (packaged) Catch. I added support for both Catch2 and Catch, and testing in TravisCI.

This is based on #286 due to eventual conflicts.

### DIFF
--- a/ci_build.sh
+++ b/ci_build.sh
@@ -8,6 +8,7 @@ ZMQ_VERSION=${ZMQ_VERSION:-4.2.5}
 ENABLE_DRAFTS=${ENABLE_DRAFTS:-OFF}
 COVERAGE=${COVERAGE:-OFF}
 LIBZMQ=${PWD}/libzmq-build
+CATCH2=${PWD}/catch2-build
 CPPZMQ=${PWD}/cppzmq-build
 # Travis machines have 2 cores
 JOBS=2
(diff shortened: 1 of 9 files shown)


QuLogic:
Related to #280.

tests/message.cpp
```diff
@@ -1,5 +1,9 @@
 #define CATCH_CONFIG_MAIN
+#ifdef USE_EXTERNAL_CATCH2
```
kurdybacha: This can be avoided as in both cases (catch2, catch1) header name is the same. Instead doing ifdef in the code you can just control include paths from cmake.

ci_build.sh
```diff
@@ -33,6 +34,18 @@ libzmq_install() {
     fi
 }
 
+catch2_install() {
+    curl -L "https://github.com/catchorg/Catch2/archive/v${CATCH2_VERSION}.tar.gz" \
```
kurdybacha: how about downloading only header file? do we need to build and install it?

QuLogic:
I do not see how using Catch2 and C++11 can be an issue. Catch2 is only used while compiling tests, and if the user has Catch2 then they have a C++11 compiler (I'm not convinced that is a hard requirement either since I was able to install and test against Catch2 with g++ 4 in trusty.) Once cppzmq is installed, then the tests are irrelevant and since it's a header-only library the C++11 requirement is gone, unless the flags get encoded in the CMake file somehow?

tests/message.cpp
sigiesec: I think it is better to stick with the include base path a library expects/defines, and I also think the #ifdef is fine in general. However, please confine this to a single point instead of duplicating this across all test files.

sigiesec:
@QuLogic Any update on this?

I think it is good to add support for using a catch provided by the user/system. After looking at the Breaking Changes section of https://github.com/catchorg/Catch2/releases/tag/v2.0.1, I am not sure if it really feasible to allow either catch or catch2. It might work right now, because we only use a very limited set of features of catch, but this might break or get tedious in the future.

I would suggest to just add support for an existing catch, and we keep in mind to update to catch if and when we drop pre-C++11 support in cppzmq.

Another alternative might be to just add the catch header to the repository, to resolve the requirement for network access. This wouldn't be feasible for a large library, but provided it is only a single header file, it might be acceptable.

gummif:
cppzmq now uses cmake fetch content for catch2 and testing can be disabled with CPPZMQ_BUILD_TESTS, so closing this now.

QuLogic:
> cppzmq now uses cmake fetch content for catch2

Yes, this is sufficient, as CMake finds Catch2 if it's installed system-wide without downloading anything.

PR #289, opened 2019-01-30
Ensure ptr is set again to NULL once the socket is disconnected

anebot:
I've just recap that once you call `Disconnect()` from your socket `ptr` is not set again to `NULL`, so a subsequent call to connected() it will return true regardless the socket is really disconnected.

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -675,10 +675,8 @@ class socket_t
     inline void disconnect(const char *addr_)
     {
         int rc = zmq_disconnect(ptr, addr_);
-        if (rc != 0) {
+        if (rc != 0)
             throw error_t();
-	}
-	ptr = NULL;
     }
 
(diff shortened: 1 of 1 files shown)


sigiesec:
Thanks for the PR. I fear this change ia not valid. The behaviour of connect/disconnect is different: You can connect to multiple endpoints, disconnect from some and connect again, for example. Apparently there are no tests for this.

And you're right, connected is names misleadingly. A better name would be valid or opened, since calling close will make a socket invalid.

sigiesec:
The underlying libzmq doesn't allow to query the list of connected endpoints, so this would need to be done on the cppzmq, which is not quite in line with the "thin layer" approach of cppzmq. Also, be aware that "connected" in the terms of libzmq doesn't mean there is an existing network (e.g. TCP) connection, just that it could connect if the peer is available.

sigiesec:
@anebot I am closing this PR, since all changes are reverted.

PR #295, opened 2019-03-08
Add message_t::get()

### LINKED ISSUE #277: zmq_msg_get not available
Hello,

Today I needed to use zmq_msg_get from the zmq C API. Unfortunately, there are:

- No way to retrieve the underlying zmq_msg_t from a zmq::message_t.
- No zmq::message_t::get method.

I had to patch the zmq.hpp by myself:

```
    inline int get(int property_)
    {
        int value = zmq_msg_get(&msg, property_);
        if (value == -1)
            throw error_t();
        return value;
    }
```

Would it be possible to add this to the mainstream?

Thanks,

trya:
Uses zmq_msg_get() to get some integer properties on the message.

Fixes #277.

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -385,6 +385,16 @@ class message_t
         return !(*this == other);
     }
 
+#if ZMQ_VERSION >= ZMQ_MAKE_VERSION(3, 2, 0)
+    inline int get(int property_)
+    {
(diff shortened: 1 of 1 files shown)


sigiesec:
@trya Thanks for the PR! Could you perhaps also add a unit test case for the new method?

lp35:
Thanks for the PR!

PR #296, opened 2019-03-13
Removed warnings when compiling with C++11 enabled

ivan-cukic:
- Replaced NULL and 0s used as pointers with ZMQ_NULLPTR (defined as nullptr on C++11, as 0 otherwise)
- Added override to the when() member function

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -42,7 +42,7 @@
 #define ZMQ_NULLPTR nullptr
 #else
 #define ZMQ_CPP03
-#define ZMQ_NOTHROW
+#define ZMQ_NOTHROW throw()
 #define ZMQ_EXPLICIT
 #define ZMQ_OVERRIDE
 #define ZMQ_NULLPTR 0
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -39,11 +39,13 @@
 #define ZMQ_NOTHROW noexcept
 #define ZMQ_EXPLICIT explicit
 #define ZMQ_OVERRIDE override
+#define ZMQ_NULLPTR nullptr
 #else
 #define ZMQ_CPP03
 #define ZMQ_NOTHROW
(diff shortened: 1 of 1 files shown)


zmq.hpp
```diff
@@ -138,11 +140,7 @@ class error_t : public std::exception
 {
   public:
     error_t() : errnum(zmq_errno()) {}
-#ifdef ZMQ_CPP11
-    virtual const char *what() const noexcept { return zmq_strerror(errnum); }
-#else
-    virtual const char *what() const throw() { return zmq_strerror(errnum); }
-#endif
+    virtual const char *what() const ZMQ_NOTHROW ZMQ_OVERRIDE { return zmq_strerror(errnum); }
```
sigiesec: This changes the signature of what for pre-C++11, since ZMQ_NOTHROW is not defined as throw() but as nothing.

I am not sure if ZMQ_NOTHROW can be defined as throw() (I don't know at which other places the macro is used as of now), or if a specific macro should be defined for this case.

It is a good idea to remove the duplication here though.

sigiesec: Actually, the exception specification is determined by the base class (i.e. the C++ standard), so in cannot be changed. The change breaks the build, see e.g. here: https://travis-ci.org/zeromq/cppzmq/jobs/505687621#L1401

ivan-cukic-matf: Fair enough. I'll try to replace the macro with `throw()` and see what happens.

ivan-cukic: Fair enough.

Other usages are for non-virtual memfns. And most of them inline.

I'll try to replace the macro with throw() and see whether Travis complains.

sigiesec: Ok, should be fine to add the nothrow declarations to these methods. Build is fine now. Thanks!

PR #297, opened 2019-03-14
Unit tests for message_t::get()

trya:
Testing ZMQ_SHARED property only, since other properties (ZMQ_MORE and ZMQ_SRCFD) depend on socket operations.

Following PR #295.


sigiesec:
Thanks a lot :)

PR #298, opened 2019-03-26
Problem: poll is mutating by definition

gummif:
zmq_poll mutates the input array, so have const qualifiers for the poll functions does not make sense.

Solution: Remove const qualifiers and const_cast

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -147,29 +147,54 @@ class error_t : public std::exception
     int errnum;
 };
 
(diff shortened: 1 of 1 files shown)


kurdybacha:
I agree with the changes but that unfortunately breaks API compatibility so we can not change it right now. We can add them as a separate functions though.

gummif:
That's true. I added overloads taking non-const items, and make the const functions deprecated. Hopefully the deprecation message is alright.

sigiesec:
Thanks :)

PR #301, opened 2019-03-30
Problem: proxy is not typesafe

gummif:
Solution: Add overloads to proxy and proxy_steerable taking socket_t
objects
Additionally deprecate the void* overloads.

### DIFF
--- a/tests/socket.cpp
+++ b/tests/socket.cpp
@@ -1,5 +1,8 @@
 #include <catch.hpp>
 #include <zmq.hpp>
+#ifdef ZMQ_CPP11
+#include <future>
+#endif
 
 TEST_CASE("socket create destroy", "[socket]")
 {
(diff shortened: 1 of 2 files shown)


tests/socket.cpp
```diff
@@ -27,3 +30,43 @@ TEST_CASE("socket sends and receives const buffer", "[socket]")
...
+    zmq::socket_t back(context, ZMQ_ROUTER);
+    zmq::socket_t capture(context, ZMQ_DEALER);
+    front.bind("inproc://test1");
+    back.bind("inproc://test2");
+    capture.bind("inproc://test3");
+    auto f = std::async(std::launch::async, [&]() {
+        auto s1 = std::move(front);
+        auto s2 = std::move(back);
+        auto s3 = std::move(capture);
+        CHECK_THROWS(zmq::proxy(s1, s2, &s3));
```
sigiesec: Does this throw because of an ETERM error because the context is closed?

Shouldn't the CHECK_THROW be placed where the future is retrieved with get?

gummif: Added explicit check for ETERM (constant available on windows?) and moved the check.

sigiesec:
Yes, it is defined under Windows as well!

Thanks for the amendment :)

PR #303, opened 2019-04-02
Problem: Missing swap functions

gummif:
Solution: Implement for socket_t, context_t, message_t and poller_t
Additionally remove dependency on `<functional>` by refactoring poller_t
and remove unused `<unordered_map>` include.

### DIFF
--- a/tests/context.cpp
+++ b/tests/context.cpp
@@ -1,6 +1,11 @@
 #include <catch.hpp>
 #include <zmq.hpp>
 
+#if (__cplusplus >= 201703L)
+static_assert(std::is_nothrow_swappable<zmq::context_t>::value,
+              "context_t should be nothrow swappable");
+#endif
+
 TEST_CASE("context construct default and destroy", "[context]")
 {
     zmq::context_t context;
@@ -12,3 +17,13 @@ TEST_CASE("context create, close and destroy", "[context]")
     context.close();
     CHECK(NULL == (void *) context);
 }
+
+#ifdef ZMQ_CPP11
+TEST_CASE("context swap", "[context]")
+{
+    zmq::context_t context1;
+    zmq::context_t context2;
+    using std::swap;
+    swap(context1, context2);
+}
+#endif
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -73,8 +73,6 @@
 #ifdef ZMQ_CPP11
 #include <chrono>
 #include <tuple>
-#include <functional>
-#include <unordered_map>
 #include <memory>
 #endif
 
@@ -476,6 +474,12 @@ class message_t
         return os.str();
     }
 
+    void swap(message_t &other) ZMQ_NOTHROW
+    {
+        // this assumes zmq::msg_t from libzmq is trivially relocatable
+        std::swap(msg, other.msg);
+    }
+
   private:
     //  The underlying message
     zmq_msg_t msg;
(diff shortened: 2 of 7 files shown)


zmq.hpp
```diff
@@ -484,8 +487,19 @@ class message_t
     //  messages (less efficient) without being aware of the fact.
     message_t(const message_t &) ZMQ_DELETED_FUNCTION;
     void operator=(const message_t &) ZMQ_DELETED_FUNCTION;
+
+    #ifdef ZMQ_CPP11
+    // zmq::msg_t must also be trivially copyable
+    static_assert(std::is_trivially_copyable<zmq_msg_t>::value,
```
sigiesec: Do you really mean zmq_msg_t here? zmq_msg_t is defined by libzmq, which has a C API, where all types are trivially copyable. Apart from that, it is not needed for std::swap to be well-defined, since that only requires move-constructability and move-assignability (which are implied by, but not necessary for trival copyability). So if anything, this should be asserted.

In any case, I would suggest to move this to the tests as well, such that it does not need to be compiled for code using this.

gummif: That's true, what I really meant to say is that the underlying type (`zmq::msg_t`) needs to be [trivially relocatable](https://quuxplusone.github.io/blog/2018/07/18/announcing-trivially-relocatable/). The C API includes `zmq_msg_move` and `zmq_msg_copy` for moving and copying messages, so it is not obvious (to me at least) that one can simply swap the C structs and everything remains OK. But after inspecting `zmq::msg_t` it appears to be quite trivial but I don't know if that is guaranteed to remain so by the API.

sigiesec: When thinking about this again, I think this might be done differently.

Yesterday, I was only thinking about the "technical" properties of zmq_msg_t in terms of the C++ concepts. However, since you mention zmq_msg_move/zmq_msg_copy, this is not equivalent to the logical properties. You cannot memcpy a zmq_msg_t (in general) and then continue to use both copies, this could lead to a double-free, e.g., so "logically", zmq_msg_t is not trivially-move-constructible/assignable, but trivially relocatable (thanks for the link, I didn't know about this concept yet, but it makes perfect sense :) ). The only question is if this is a property of the libzmq API, or an implementation detail, which might change.

Providing an explicit swap specialization for a type is only useful if the type is either non-move-assignable or non-move-constructible, or the swapping can be done more efficiently than the default swap, or you are on pre-C++11, which you seem to want to support here, right? Otherwise, the default swap implementation could probably be used.

What I am wondering is why the move constructor and assignment operator of message_t are not implemented using zmq_msg_move. I think this would be nicer, if not required.

tests/context.cpp
```diff
@@ -12,3 +17,13 @@ TEST_CASE("context create, close and destroy", "[context]")
...
     CHECK(NULL == (void *) context);
 }
+
+#ifdef ZMQ_CPP11
+TEST_CASE("context swap", "[context]")
+{
+    zmq::context_t context1;
+    zmq::context_t context2;
+    swap(context1, context2); // found by adl
+    zmq::swap(context1, context2);
```
sigiesec: Should this really be tested? Isn't it sufficient, if it works like this:
```
using std::swap;
swap(context1, context2);
```
This will work either by using a default overload from std or a custom overload using ADL from zmq.

zmq.hpp
gummif: Would you want another pull request that changes message_t to use zmq_msg_move where needed (move ctor, move assign, swap), even if that might not be as fast as copying structs, to be on the safe side?

sigiesec: Yes, if you want to do that, that would be great. 

Apart from the fact that, if libzmq is used as a shared library, this will involve an extra function call that cannot be inlined, the current implementation of zmq_msg_move will more or less be identical to the current implementation of the move ctor and assignment operator.

PR #306, opened 2019-04-03
Problem: Missing detection of standards greater than C++11

gummif:
Solution: Add macros detecting C++14 and C++17

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -26,7 +26,18 @@
 #ifndef __ZMQ_HPP_INCLUDED__
 #define __ZMQ_HPP_INCLUDED__
 
-#if (__cplusplus >= 201402L)
+// macros defined if has a specific standard or greater
+#if (defined(__cplusplus) && __cplusplus >= 201103L) || (defined(_MSC_VER) && _MSC_VER >= 1900)
+    #define ZMQ_CPP11
+#endif
+#if (defined(__cplusplus) && __cplusplus >= 201402L) || (defined(_HAS_CXX14) && _HAS_CXX14 == 1)
+    #define ZMQ_CPP14
+#endif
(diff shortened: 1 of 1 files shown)


zmq.hpp
```diff
@@ -26,22 +26,37 @@
...
 #elif defined(_MSC_VER)
 #define ZMQ_DEPRECATED(msg) __declspec(deprecated(msg))
 #elif defined(__GNUC__)
 #define ZMQ_DEPRECATED(msg) __attribute__((deprecated(msg)))
 #endif
 
-#if (__cplusplus >= 201103L) || (defined(_MSC_VER) && (_MSC_VER >= 1900))
-#define ZMQ_CPP11
+#if defined(ZMQ_CPP17)
+#define ZMQ_NODISCARD [[nodiscard]]
```
sigiesec: This adds a definition of ZMQ_NODISCARD, but it this macro is not used anywhere. Are you working on a adding uses of this to zmq.hpp?

gummif: Yes, `bool empty()` for message_t would be useful, where nodiscard would be appropriate.

sigiesec:
Sure. I think there are many other methods as well where nodiscard would make sense. 

My question was if you intend to add these attributes in another PR soon, so we don't have a state with this unused definition around for long.

zmq.hpp
sigiesec: Sure. I think there are many other methods as well where nodiscard would make sense.

My question was if you intend to add these attributes in another PR soon, so we don't have a state with this unused definition around for long.

gummif: Yes, I could do that within a couple of days.

sigiesec: Great, I will merge this PR now.

PR #307, opened 2019-04-04
Problem: poller_t does not need to be a template class

gummif:
Solution: Remove template, and make the add function a template instead.

The template parameter of poller_t as it is does not really add any type safety, so we can just drop it.

### DIFF
--- a/tests/poller.cpp
+++ b/tests/poller.cpp
@@ -6,37 +6,36 @@
 #include <memory>
 
 #if (__cplusplus >= 201703L)
-static_assert(std::is_nothrow_swappable<zmq::poller_t<>>::value,
-              "poller_t should be nothrow swappable");
+static_assert(std::is_nothrow_swappable_v<zmq::poller_t>);
 #endif
 
 TEST_CASE("poller create destroy", "[poller]")
 {
-    zmq::poller_t<> poller;
+    zmq::poller_t poller;
 }
 
-static_assert(!std::is_copy_constructible<zmq::poller_t<>>::value,
+static_assert(!std::is_copy_constructible<zmq::poller_t>::value,
               "poller_t should not be copy-constructible");
-static_assert(!std::is_copy_assignable<zmq::poller_t<>>::value,
+static_assert(!std::is_copy_assignable<zmq::poller_t>::value,
               "poller_t should not be copy-assignable");
 
 TEST_CASE("poller move construct empty", "[poller]")
 {
-    zmq::poller_t<> a;
-    zmq::poller_t<> b = std::move(a);
+    zmq::poller_t a;
+    zmq::poller_t b = std::move(a);
 }
 
 TEST_CASE("poller move assign empty", "[poller]")
 {
-    zmq::poller_t<> a;
-    zmq::poller_t<> b;
+    zmq::poller_t a;
+    zmq::poller_t b;
     b = std::move(a);
 }
 
 TEST_CASE("poller swap", "[poller]")
 {
-    zmq::poller_t<> a;
-    zmq::poller_t<> b;
+    zmq::poller_t a;
+    zmq::poller_t b;
     using std::swap;
     swap(a, b);
 }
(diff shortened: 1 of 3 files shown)


kurdybacha:
The code looks cleaner after this change.
Currently you will get compilation error if you try to pass pointer to something different than your template argument type. That prevents passing different types to `user_data` as there is no clean way (except mapping `socket_t*` to `user_data` type to find out what you have in `user_data` while iterating through ready events but that is ok I think and it keeps it more in sync with libzmq.

sigiesec:
The template argument is present on poller_t exactly to prevent passing different types as user data in different calls. The default void argument value is only present for backwards compatibility, since using void* is not very C++-like. I do not think this should be changed.

Even if the template argument were removed from poller_t, making add a template does not make sense, since that does not provide any benefits over having an add accepting void*.

kurdybacha:
@sigiesec looking from user perspective maybe we should not prevent user to pass different types. Why limit this scenario for the users? In this case template makes more sense over void* because user does not have to reinterpret_cast anything to void* and code on the client side is cleaner.

sigiesec:
Casting any pointer to void* is implicit.

It is potentially dangerous to pass different types. A design goal of cppzmq is to provide a C++ idiomatic API without requiring the use of void*. Unless there is a very good reason to allow this, I wouldn't.

Can you write or at least sketch a test case where adding unrelated types would make sense? (Not just syntactically possible, but one where this looks like a better alternative than passing the same type).

gummif:
Thanks for the comments. @sigiesec you have a valid point that the template of add does not add any safety over just `void*`. And I don't think I can come up with a good example where having different types of `user_data` is a good design by default

However I think if we keep `poller_t` as a template class, then we should improve the API by making `wait_all` take a type depeding on `T`. That requires a template based on `zmq_poller_event_t`
```cpp
namespace zmq
{
struct no_user_data;

template<class T = no_user_data>
struct poller_event_t
{
    void *socket; // keep this as void* ?
#if defined _WIN32
    SOCKET fd;
#else
    int fd;
#endif
    T *user_data; // was void*
    short events;
};
}
```
A side note: having a non-owning `socket_view` type would be useful here instead of `void*` (where `sizeof(socket_view) == sizeof(void*)`). The `void *socket` in that struct is very incompatible with the cppzmq API as you can not use it in any way except comparing it to existing `zmq::socket_t` handles. Thoughts on this?

Now poller_t could look something like
```cpp
template<class T = no_user_data>
class poller_t
{
public:
    //...

    /* only enabled if T is not no_user_data, something like*/
    template<typename = typename std::enable_if<!std::is_same<T, no_user_data>::value>::type>
    void add(zmq::socket_t &socket, short events, T *user_data = nullptr)
    {
        add_impl(socket, events, user_data)
    }

    void add(zmq::socket_t &socket, short events)
    {
        add_impl(socket, events, nullptr);
    }

    // poller_events are now type safe and do not require casts from void*
    size_t wait_all(std::vector<poller_event_t<T>> &poller_events,
                    const std::chrono::microseconds timeout)
    {
        // use reinterpret_cast<zmq_poller_event_t*>(poller_events.data())
    }
    //...
};
```

The default of `poller_t<>` would take no user data and calling `add` with a `user_data` pointer is not enabled, for any user data you would explicitly say `poller_t<void>`, and for a poller with callbacks you could say `poller_t<void(int)>`.

sigiesec:
@gummif Sounds good to me overall. Since this is in the DRAFT part, it is fine to change it incompatibly. 

Just one note: you don't need to use reinterpret_cast to cast from void* to T*, static_cast is fine for that, and avoids warnings by static analysis tools such as clang-tidy.

gummif:
OK. I think a valuable first step would be to introduce a new `zmq::socket_ref` type, a light-weight non-owning nullable reference type compatible with the layout of void*. I think the basic design can be based on the future [std::function_ref](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0792r0.html) and `std::string_view`.
With that we can hopefully modernize and un-void*-ify the API a bit.

gummif:
Closing because problem can be solved with a bit more ambitious plan towards type safety.

PR #309, opened 2019-04-07
Problem: zmq_poller_wait_all is called with microseconds count

gummif:
Solution: Change wait_all parameter type to std::chrono::milliseconds
"zmq_poller_wait_all shall wait timeout milliseconds" http://api.zeromq.org/4-3:zmq-poller


sigiesec:
Oops, this was a bug that would probably never have been found by a test! Thanks for looking so thoroughly!

PR #310, opened 2019-04-09
Problem: catch.hpp included in coverage

gummif:
Solution: Exclude build directory for coveralls


gummif:
coveralls wont update, closing and will try opening another PR.

PR #311, opened 2019-04-09
Problem: catch.hpp included in coverage

gummif:
Solution: Exclude file for coveralls


gummif:
coveralls are still not updating, not sure of there is a problem with this PR or coveralls

sigiesec:
Hm, I can give it a look later on. I find the behaviour of coveralls somewhat hard to understand, compared to Bullseye Coverage, e.g., but that's not freely available, unfortunately.

gummif:
Fixed in another PR

Issue #140, opened 2017-07-31
C++ thread API like native_handle


ovanes:
I think it'd be great to have instead of `operator void*()` to have `native_handle()` as in thread API:

http://en.cppreference.com/w/cpp/thread/thread/native_handle

I would like to go consequently about usage of `auto`, but not having a function here causes either to cast, use explicit type `void*` or to call the `context.operator void*()`.

Issue #84, opened 2016-08-26
EAGAIN should also be thrown


rafaelroquetto:
Hello,

perhaps I am missing the rationale behind the decision to return 0 (or false) inside recv and send when an EAGAIN is spotted, instead of throwing - and I apologize in advance if this is the case.

The problem I encountered with this approach is that, when returning 0 only (instead of explicitly throwing EAGAIN like the other errors), I cannot tell, specially in the context of recv, whether I am reading a valid message with an empty frame (with size 0), or if this is an actual case of EAGAIN.

Is there a way to tell apart a genuine EAGAIN error versus reading an empty frame of a multipart message? Or should EAGAIN really be thrown?

Thanks,
Rafael

dgsteffen:
Hi -- I just ran into the same problem, but the other way around. :-) I misread the documentation and was expecting a return value of -1 if something went wrong.  But, I wasn't expecting exceptions to be thrown -- the text about "will throw exceptions" is elsewhere in the documentation.

I think the interface is fine -- return true if you got a message, return false if you didn't, throw if something really went wrong.  However, I haven't fiddled with multipart messages, so I don't know if there's something else we need to do there.

However, I'd recommend adding some details to the documentation.  For example, under "bool socket_t::recv(message_t *msg, int flags = 0)"  I'd put the "maps to zmq_recv" information last, since that's the least important thing.  Put "Returns true if..." part first, and add a statement "throws on error condition" there.

I'll submit a patch against the docs if you'll point me to where they are.... :-)

MirrorSaver:
I want to add to this question. I was under the assumption, that recv would throw in nonblocking mode if nothing is in the queue.
```
try
{
    while (true)
    {

      recv = skt_recv.recv(&zmq_topic,ZMQ_NOBLOCK);
      skt_recv.getsockopt(ZMQ_RCVMORE, &more, &more_size)
      while (more)
      {
           recv3 = skt_recv.recv(&zmq_overtaking_status,ZMQ_NOBLOCK);
           skt_recv.getsockopt(ZMQ_RCVMORE, &more, &more_size);
      }
    }
}
catch(const std::exception &)
{}
```

I want to always empty the queue so I have the newest message, even if the sender might be faster. However without the throwing how can I decide if its just nothing published or if i emptied the whole queue?

Issue #61, opened 2015-11-16
send has ambiguous error/success return


oblitum:
```c++
        inline size_t send (const void *buf_, size_t len_, int flags_ = 0)
        {
            int nbytes = zmq_send (ptr, buf_, len_, flags_);
            if (nbytes >= 0)
                return (size_t) nbytes;
            if (zmq_errno () == EAGAIN)
                return 0;
            throw error_t ();
        }
```

For `nbytes == 0` (e.g. parts for size zero), the return is zero, but it's success, nonetheless, a zero return is being used to mean EAGAIN.

mwpowellhtx:
Why are you calling send when you know your buffer doesn't have anything to send?

oblitum:
@mwpowellhtx my particular case relates with sending file contents as a message-part, and I didn't bother to have additional logic for special messages when files are empty, it should have the same logic. I guess it shouldn't have special treatment and it seems ZMQ raw api supports this since it allows for zero return with success and reserves -1 return to mean any error.

xaqq:
Hello,

I agree that the behavior is not great, especially since it is not coherent with the `bool send()` function, which returns `false` on `EAGAIN` and `true` on 0 bytes sent.

I am not sure however what the appropriate fix would be:
  + Changing the return type to `int`, and return `-1` on `EAGAIN`.
  + Throw on `EAGAIN` (I don't like this)
  + Any idea?

maxlein:
I have also a question in this direction:
Dealer Socket which is not connected sends a message out. 
Result:
I get "40 bytes send" back and the function returns true, although an error is set. This is somehow misleading.

nbytes: 40, errno: Resource temporarily unavailable

` inline bool send(message_t &msg_, int flags_ = 0)
    {
        int nbytes = zmq_msg_send(&(msg_.msg), ptr, flags_);
        std::cerr << "nbytes: " << nbytes << ", errno: " << zmq_strerror (zmq_errno()) << std::endl;
        if (nbytes >= 0)
            return true;
        if (zmq_errno() == EAGAIN)
            return false;
        throw error_t();
    }`

PR #312, opened 2019-04-10
 Problem: No type-safe alternatives when polling or needing a reference to a socket

gummif:
Solution: Introduce a socket_ref that is a non-owning nullable reference to a socket

This is a quite big PR with a few small additional types to improve type safety. The API is intentionally minimal, but can be extended at later points. Some design decision:

1. `socket_ref` is a trivially copyable type with the same layout as `void*`, making it a step towards typesafe polling
2. `socket_ref` is provides comparison and hashing, making it suitable for containers like `std::map` or `std::unordered_set`. In addition to being a non-owning reference, makes the use of `void*` obsolete (for C++11).
3. A `handle()` function is provided to get the underlying `void*` (solving https://github.com/zeromq/cppzmq/issues/140), since `native_handle()` implies an operating system native socket handle, the `handle()` name was chosen instead
3. A function taking void* is quite greedy, so to construct `socket_ref` from an existing handle one must use the `zmq::from_handle` utility to be explicit
4. Flag enums for safer `send` and `recv`
5. Result structs from `send` and `recv` removing ambiguity on EAGAIN (solving https://github.com/zeromq/cppzmq/issues/84 and https://github.com/zeromq/cppzmq/issues/61)

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -56,11 +56,15 @@
 #define ZMQ_EXPLICIT explicit
 #define ZMQ_OVERRIDE override
 #define ZMQ_NULLPTR nullptr
+#define ZMQ_CONSTEXPR_FN constexpr
+#define ZMQ_CONSTEXPR_VAR constexpr
 #else
 #define ZMQ_NOTHROW throw()
 #define ZMQ_EXPLICIT
 #define ZMQ_OVERRIDE
 #define ZMQ_NULLPTR 0
+#define ZMQ_CONSTEXPR_FN
+#define ZMQ_CONSTEXPR_VAR const
 #endif
 
 #include <zmq.h>
@@ -230,8 +234,6 @@ inline std::tuple<int, int, int> version()
 
 class message_t
 {
-    friend class socket_t;
-
   public:
     message_t()
     {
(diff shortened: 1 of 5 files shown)


zmq.hpp
```diff
@@ -634,7 +634,146 @@ enum class socket_type : int
...
+{
+    struct _private {}; // disabling use other than with from_handle
+    constexpr explicit from_handle_t(_private /*p*/) noexcept {}
+};
+
+constexpr from_handle_t from_handle = from_handle_t{from_handle_t::_private{}};
+
+// A non-owning nullable reference to a socket.
+// The reference is invalidated on socket close or destruction.
+class socket_ref
```
sigiesec: I am a bit confused on the relationship of socket_ref and socket. So now socket_ref has new send/recv variants, which are not present on socket. This is somewhat confusing IMO.

However, I am not sure what I would expect from the API of a socket_ref. Wouldn't one want to have the same operations as on socket?

That would lead to quite a lot of duplication though.

Wouldn't it be easier and implicitly consistent to use the same class for an owning and a non-owning socket, or at least the same base class (or base class template)?

zmq.hpp
```diff
@@ -634,7 +634,146 @@ enum class socket_type : int
...
+}
+inline bool operator!=(socket_ref a, socket_ref b) noexcept
+{
+    return !(a == b);
+}
+inline bool operator<(socket_ref a, socket_ref b) noexcept
+{
+    return a.handle() < b.handle();
+}
+inline bool operator>(socket_ref a, socket_ref b) noexcept
```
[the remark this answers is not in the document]
gummif: For completeness, but arguable not very useful. Would you like to see only < implemented?

zmq.hpp
gummif: Yes, the API is supposed to be mostly the same except for the `close` function. `socket_t` also has the burden of pre-C++11 support (is dropping that support on the horizon?), but I will see if I can refactor the common functionality into a base class.

gummif:
I have moved common functionality into `socket_base`, made `socket_ref` compile with C++98, replaced uses of `void*` to `socket_ref` and `socket_t` where possible, and added a `handle()` function to `message_t` removing the need for friend. The `send` and `recv` variants previously in the PR will be the subject of another PR.

sigiesec:
Thanks for the update, I was AFAIK for the last days. I hope I can give it a look later today.

sigiesec:
Sorry it took so long, I now merged this. Thanks again for the PR and the followup changes!

PR #315, opened 2019-04-14
Problem: Compiler warnings in tests

gummif:
Solution: Fix some compiler warnings

### DIFF
--- a/tests/poller.cpp
+++ b/tests/poller.cpp
@@ -87,7 +87,7 @@ TEST_CASE("poller add handler invalid events type", "[poller]")
     zmq::socket_t socket{context, zmq::socket_type::router};
     zmq::poller_t<> poller;
     short invalid_events_type = 2 << 10;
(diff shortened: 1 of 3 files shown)


tests/poller.cpp
```diff
@@ -107,7 +107,7 @@ TEST_CASE("poller wait with no handlers throws", "[poller]")
     std::vector<zmq_poller_event_t> events;
     /// \todo the actual error code should be checked
     CHECK_THROWS_AS(poller.wait_all(events, std::chrono::milliseconds{10}),
-                 zmq::error_t);
+                 const zmq::error_t&);
```
sigiesec: Why is it necessary to add this?

According to https://github.com/catchorg/Catch2/blob/master/docs/assertions.md:
> Note that the exception type is extended with const& and you should not include it yourself.

gummif: This was not the case in v1.x of Catch (that's the version we are using?), and gcc gave me conversion warnings.

sigiesec: Ah yes, you're right. I didn't check the docs of Catch 1.x.

PR #316, opened 2019-04-23
Use C++ cast instead of old style cast

texierp:
This to avoid the following warning with the '-Wold-style-cast' flag enabled :

./zmq.hpp:763:29: warning: use of old-style cast [-Wold-style-cast]
             return (size_t) nbytes;
                             ^
./zmq.hpp: In member function ‘size_t zmq::socket_t::recv(void*, size_t, int)’:
./zmq.hpp:793:29: warning: use of old-style cast [-Wold-style-cast]
             return (size_t) nbytes;


sigiesec:
Thanks for the PR :)

PR #317, opened 2019-05-02
Problem: send/recv functions lack type-safety

gummif:
Solution: Add functions taking buffers and enum class flags

1. `enum class` flags `send_flags` and `recv_flags` instead of `int`s for safer `send` and `recv`.
2. Result structs `send_result`, `recv_result` and `recv_buffer_result` so that no information returned from libzmq is lost, removing ambiguity on EAGAIN (solving https://github.com/zeromq/cppzmq/issues/84 and https://github.com/zeromq/cppzmq/issues/61).
3. Buffer classes replacing `void*` and `size_t` pairs, based on `boost::asio::buffer` and the [Networking TS](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/n4771.pdf) `std::net::buffer` that will likely get added to C++23. This synchronizes the API of cppzmq with future standard C++ networking API (and current `boost::asio`), making it easier to use (and harder to misuse) and more familiar for newcomers. 

0MQ does not support sending or receiving buffer sequences (that would be more equivalent to multipart, which should be explicit anyway), so `send` and `recv` just take `const_buffer` and `mutable_buffer` (and not `ConstBufferSequence` and `MutableBufferSequence` as asio does). I envision `send_multipart` and `recv_multipart` taking ranges of `message_t`s or `buffer`s will make the library API more powerful, efficient and consistent, making the `multipart_t` container unnecessary.

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -280,7 +280,7 @@ class message_t
     }
 
 #if defined(ZMQ_BUILD_DRAFT_API) && defined(ZMQ_CPP11)
-    // this function is too greedy, must add
+    // TODO: this function is too greedy, must add
     // SFINAE for begin and end support.
     template<typename T>
     explicit message_t(const T &msg_) : message_t(std::begin(msg_), std::end(msg_))
@@ -654,7 +654,8 @@ enum class send_flags : int
 
 constexpr send_flags operator|(send_flags a, send_flags b) noexcept
 {
-    return static_cast<send_flags>(static_cast<int>(a) | static_cast<int>(b));
+    return static_cast<send_flags>(static_cast<std::underlying_type<send_flags>::type>(a)
+        | static_cast<<std::underlying_type<send_flags>::type>(b));
 }
 
 enum class recv_flags : int
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -72,16 +72,21 @@
 #include <cassert>
 #include <cstring>
 
-#ifdef ZMQ_CPP11
-#include <array>
-#endif
 #include <algorithm>
 #include <exception>
 #include <iomanip>
-#include <iterator>
 #include <sstream>
 #include <string>
 #include <vector>
+#ifdef ZMQ_CPP11
+#include <array>
+#include <chrono>
+#include <tuple>
+#include <memory>
+#endif
+#ifdef ZMQ_CPP17
+#include <optional>
+#endif
 
 /*  Version macros for compile-time API version detection                     */
 #define CPPZMQ_VERSION_MAJOR 4
@@ -91,12 +96,6 @@
 #define CPPZMQ_VERSION                                                              \
     ZMQ_MAKE_VERSION(CPPZMQ_VERSION_MAJOR, CPPZMQ_VERSION_MINOR,                    \
                      CPPZMQ_VERSION_PATCH)
-
-#ifdef ZMQ_CPP11
-#include <chrono>
-#include <tuple>
-#include <memory>
-#endif
 
 //  Detect whether the compiler supports C++11 rvalue references.
 #if (defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ > 2))   \
@@ -621,23 +620,11 @@ inline void swap(context_t &a, context_t &b) ZMQ_NOTHROW {
 }
 
 #ifdef ZMQ_CPP11
-struct send_result
-{
-    size_t size;    // message size in bytes
-    bool success;
-};
 
-struct recv_result
-{
-    size_t size;    // message size in bytes
-    bool success;
-};
-
-struct recv_buffer_result
+struct recv_buffer_size
 {
     size_t size;    // number of bytes written to buffer
     size_t untruncated_size;  // untruncated message size in bytes
-    bool success;
 
     ZMQ_NODISCARD bool truncated() const noexcept
     {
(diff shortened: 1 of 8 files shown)


gummif:
A min macro appears to be defined on windows causing failures, will try to find a way around that.

gummif:
One thing about this design I'm not sure about is whether the result structs should expose its members, or provide getters like `size()` and `explicit operator bool()`.

zmq.hpp
```diff
@@ -615,8 +620,340 @@ inline void swap(context_t &a, context_t &b) ZMQ_NOTHROW {
...
+    size_t untruncated_size;  // untruncated message size in bytes
+    bool success;
+
+    ZMQ_NODISCARD bool truncated() const noexcept
+    {
+        return size != untruncated_size;
+    }
+};
+
+enum class send_flags : int
```
sigiesec: I am not 100% sure if an enum should be used for flags. 

While IIRC the C & C++ standards specifically allow this, I find it quite surprising that there are possible values that are not listed in the enum declaration. On the other hand, the enum name indicates that these are flags, so the probability of confusion would be acceptable.

However, unfortunately, there is no perfect solution for this in C++. An alternative would be to use a bitfield for this purpose. What would you think about that? 

In any case, this should be done uniformly with cppzmq, and already your other open PR also introduces a flags enum, so we should take a moment to be sure this is the way to go. It would also be good to add some documentation on such general conventions, so a future discussion can reference that. Maybe an additional markdown file CONVENTIONS.md could be added for that.

zmq.hpp
```diff
@@ -615,8 +620,340 @@ inline void swap(context_t &a, context_t &b) ZMQ_NOTHROW {
...
+{
+    none = 0,
+    dontwait = ZMQ_DONTWAIT,
+    sndmore = ZMQ_SNDMORE
+};
+
+constexpr send_flags operator|(send_flags a, send_flags b) noexcept
+{
+    return static_cast<send_flags>(static_cast<int>(a) | static_cast<int>(b));
```
sigiesec: Using std::underlying_type here would reduce redundancy (but would be more verbose):
```suggestion
    return static_cast<send_flags>(static_cast<std::underlying_type<send_flags>::type>(a)
        | static_cast<<std::underlying_type<send_flags>::type>(b));
```

zmq.hpp
```diff
@@ -615,8 +620,340 @@ inline void swap(context_t &a, context_t &b) ZMQ_NOTHROW {
...
+inline const_buffer buffer(const const_buffer& cb, size_t n) noexcept
+{
+    return const_buffer(cb.data(), (std::min)(cb.size(), n));
+}
+
 namespace detail
 {
+// utility functions for containers with data and size
+// data is nullptr if the container is empty
+template<class T> mutable_buffer buffar_mut_ds(T &data) noexcept
```
sigiesec: buffar -> buffer here and for several other functions below

I am not a big friend of abbreviations, would you consider changing the abbreviations to the full form?

I assume ds stands for "data and size", so this function "construct a mutable buffer from data and size"? 

It is not completely clear to be what you mean by "containers with data and size". Do you mean a [ContigousContainer](https://en.cppreference.com/w/cpp/named_req/ContiguousContainer). Then it would be better not to use a data member function, but the iterators, such that this only depends on what is guaranteed by the ContiguousContainer concept.

Maybe naming the methods something with "contiguous container" would be more appropriate then.

zmq.hpp
```diff
@@ -615,8 +620,340 @@ inline void swap(context_t &a, context_t &b) ZMQ_NOTHROW {
...
+template<class T, size_t N>
+const_buffer buffer(const T (&data)[N], size_t n_bytes) noexcept
+{
+    static_assert(detail::is_pod_like<T>::value, "T must be POD");
+    static_assert(N > 0, "N > 0");
+    return const_buffer(static_cast<const T *>(data),
+                        (std::min)(N * sizeof(T), n_bytes));
+}
+// std::array
+template<class T, size_t N> mutable_buffer buffer(std::array<T, N> &data) noexcept
```
sigiesec: Why are these overloads not using the functions from detail?

Is it necessary to provide separate overloads for array, vector and basic_string specifically, all of which could be implemented in the same way, as they are ContiguousContainers (see comment above)? That would be much less redundant and less verbose.

zmq.hpp
```diff
@@ -615,8 +620,340 @@ inline void swap(context_t &a, context_t &b) ZMQ_NOTHROW {
...
+}
+inline mutable_buffer operator+(size_t n, const mutable_buffer &mb) noexcept
+{
+    return mb + n;
+}
+
+class const_buffer
+{
+  public:
+    const_buffer() noexcept : _data(nullptr), _size(0) {}
```
sigiesec: The constructors and data and size member functions could be constexpr, as well as the factory functions below. (Not sure about the operators)

zmq.hpp
```diff
@@ -615,8 +620,340 @@ inline void swap(context_t &a, context_t &b) ZMQ_NOTHROW {
     a.swap(b);
 }
 
+#ifdef ZMQ_CPP11
+struct send_result
```
sigiesec: While this would restrict this to C++17, send_result and recv_result could be defined as a std::optional<size_t>. While the use of operator*/-> have UB in case no value is set, misuses would be more obvious, and probably trigger an assertion in typical standard library implementations.

zmq.hpp
gummif: The standard library has a BitmaskType requirement https://en.cppreference.com/w/cpp/named_req/BitmaskType . The latest bitmask additions to the library (`std::launch` for `std::async` and those in `std::filesystem`) are all specified as `enum class`, so I believe that convention is quite common for those using newer standards.
I could also comment in code that those flags satisfy (partially, not all bit operations are implemented) the named requirement BitmaskType, would that be satisfactory?

zmq.hpp
gummif: Agree, the name could be improved. Yes, it accepts contiguous containers or views/sequences that have data and size member functions. I don't know if there is a concept for that, but anything that could be used with `std::data` and `std::size`. Maybe `buffer_mutable_contiguous_sequence`.

zmq.hpp
gummif: I'm just following the spec as defined. I think it is like this because these are the types that are known to be contiguous in the standard library.
We could simplify this a lot by implementing one function, maybe using 
1. SFINAE for data and size (might be risky if called with some external type that happens to not be contiguous, but that may be the users fault anyway), or 
2. checking if the type is one of the known standard library types to be contiguous (a bit more verbose, but safe)

There are always the edge cases of C arrays and string::data() pre C++17 is always const.

zmq.hpp
gummif: That's true. I will see how constexpr I can make it.

zmq.hpp
sigiesec: Ok, wasn't aware of that the standard library does it like that as well. Then leave it as it is :) Maybe add the comment as suggested, but it is not essential.

zmq.hpp
gummif: The initial thought was that additional members could be added to the struct in a backwards compatible fashion, but since it is highly unlikely that the libzmq API will change I agree that that is a more robust and better solution to use `std::optional`.

We could still support C++11 by returning a "implementation defined" type (only pre-C++17) that emulates the basic operations of optional for trivial types, just `operator*`, `operator->`, `has_value()` and `operator bool`. For C++17 we just use `std::optional`.

zmq.hpp
sigiesec: You mean the Networking TS here? Ok, maybe leave it like this then. I am not sure, it might get complex to SFINAE this without C++20 concepts, and maybe that's why the Networking TS does it like this as well, and it has not yet been updated to use C++20 language features?

It was just an idea if it had reduced complexity.

zmq.hpp
sigiesec: Only if you have time, this could be added later on as well. I just noticed this since you used constexpr at some other places.

Sorry I didn't make clear in my review comments what were just ideas, and what I feel should be changed right now.

zmq.hpp
sigiesec: The existing libzmq API functions will probably never be changed. Theoretically new functions could be added that would be used in the existing cppzmq functions, but this appears rather unlikely to me as well. 

If you would like to add such a minimal std::optional drop-in for older standards, that would be great!

zmq.hpp
```diff
@@ -615,8 +620,341 @@ inline void swap(context_t &a, context_t &b) ZMQ_NOTHROW {
...
+{
+    none = 0,
+    dontwait = ZMQ_DONTWAIT,
+    sndmore = ZMQ_SNDMORE
+};
+
+constexpr send_flags operator|(send_flags a, send_flags b) noexcept
+{
+    return static_cast<send_flags>(static_cast<std::underlying_type<send_flags>::type>(a)
+        | static_cast<<std::underlying_type<send_flags>::type>(b));
```
sigiesec:
```suggestion
        | static_cast<std::underlying_type<send_flags>::type>(b));
```

gummif: Will be fixed in another commit.

gummif:
Thanks for the feedback. I think we are all set now.

PR #319, opened 2019-05-08
Problem: zmq_msg_init never fails

gummif:
Solution: Mark functions noexcept and assert instead of throwing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,4 +85,3 @@ script:
   - eval "${MATRIX_EVAL}"
   - cmake --version
   - ./ci_build.sh
-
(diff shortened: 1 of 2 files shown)


gummif:
> @gummif Thanks again for the improvements!
> 
> Just one request: Please split the commit into two with a specific commit message, since the problems addressed as well as the solutions are independent from each other. Please keep this in mind for future changes. This helps a lot when inspecting the history in the future.

No problem. Another PR has been submitted.

sigiesec:
Thanks! A separate PR is fine of course. However, I just meant a separate commit within this branch/PR.

PR #323, opened 2019-05-13
Problem: message_t ctor for iterators double initializes the message

gummif:
Solution: Remove constructor call

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -248,7 +248,7 @@ class message_t
             throw error_t();
     }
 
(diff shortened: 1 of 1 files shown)


sigiesec:
Thanks for this PR as well. The changes are not extensive, but the commit is mainly about changing the name of the template parameter, which makes sense but is unrelated to the commit message. Could you split this into two commits please?

gummif:
Sorry about that, bad habit.

PR #325, opened 2019-05-15
Problem: message_t ctor for ranges too greedy

gummif:
Solution: Detect ranges with enable_if idiom

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -156,6 +156,22 @@ namespace zmq
 #ifdef ZMQ_CPP11
 namespace detail
 {
+namespace ranges
+{
+using std::begin;
+using std::end;
+template<class T>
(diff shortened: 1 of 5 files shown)


zmq.hpp
```diff
@@ -152,6 +152,39 @@ typedef struct
...
+
+template<class T, class = void> struct is_range : std::false_type
+{
+};
+
+template<class T>
+struct is_range<
+  T,
+  void_t<decltype(
+    std::begin(std::declval<typename std::remove_reference<T>::type &>())
```
sigiesec: Isn't this too restrictive? std::begin isn't necessarily defined for every range, there might be a begin in a different namespace available via ADL. This also applies to the use in the meaaage_t ctor below.

gummif: Fixed. I expected it to be harder to implement, but turned out to be quite simple.

PR #328, opened 2019-06-01
Fix 'recv' function 'flags_' argument default value

xiphon:


sigiesec:
Thanks for the PR!

It's not completely clear to me what problem this fixes. Could you add a test case that is fixed by your change?

xiphon:
`flags_` argument was defaulted to `0` previously (`v4.3.0`), i.e. `bool recv(message_t *msg_, int flags_ = 0);` https://github.com/zeromq/cppzmq/blob/v4.3.0/zmq.hpp#L712

That was obviously accidentally changed in https://github.com/zeromq/cppzmq/pull/317/commits/3d4be814e8a6cb44e39ddc7c272bdcc1ffad296f resulting in non-optional `flags_` breaking backward compatibility for client code invoking `bool result = recv(&msg);` call.

Let me know if you have any further questions.

xiphon:
Updated. Added a test case.

gummif:
You are correct that this is a bug, and I think the fix is simply removing the `#ifdef` completely, that is `int flags_ = 0` as it was before.

xiphon:
Updated

Issue #332, opened 2019-06-12
Example doesn't compile


louwers:
When trying to compile the provided example:

```
[bart@samadhi zmqtest]$ g++ --std=c++17 main.cc 
main.cc: In function ‘int main()’:
main.cc:9:19: error: ‘buffer’ is not a member of ‘zmq’
    sock.send(zmq::buffer(m), zmq::send_flags::dontwait);
                   ^~~~~~
main.cc:9:35: error: ‘zmq::send_flags’ has not been declared
    sock.send(zmq::buffer(m), zmq::send_flags::dontwait);
[bart@samadhi zmqtest]$ cat main.cc 
#include <string>
#include <zmq.hpp>
int main()
{
   zmq::context_t ctx;
   zmq::socket_t sock(ctx, zmq::socket_type::push);
   sock.bind("inproc://test");
   const std::string_view m = "Hello, world";
   sock.send(zmq::buffer(m), zmq::send_flags::dontwait);
}

```

kurdybacha:
Hi @louwers , what gcc version?

louwers:
@kurdybacha gcc/g++ 9.1.0

sigiesec:
I cannot reproduce this with gcc 9.1.1. Are you sure the include path is correct, and you are using the matching version of zmq.hpp?

aufziehvogel:
What would be the matching version? I experience the same problem. I am quite new to C++, so instead of compiling (and according to the docs installing the compiled artifacts to my system) I tried to use the C++ bindings shipped by the Arch Linux package manager. They are supposed to be version 4.3.1.

```
community/zeromq 4.3.1-2
    Fast messaging system built on sockets. C and C++ bindings. aka 0MQ, ZMQ.
```

aufziehvogel:
Just checked the code files a bit. Indeed, the Arch provided header does not include `send_flags`.

It's between these two commits that it was added:
- https://github.com/zeromq/cppzmq/blob/95d519fade9c43419f6fb4ade4d390c483dacf64/zmq.hpp
- https://github.com/zeromq/cppzmq/blob/0ce8ef06d543288fa092745e399e914c1067a895/zmq.hpp

[...]

PR #338, opened 2019-08-15
Clarify example requirements

gummif:
See issue #332

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,16 +41,17 @@ Supported platforms
 
 Examples
 ========
+This example requires at least C++11.
 ```c++
 #include <string>
 #include <zmq.hpp>
+
 int main()
 {
-   zmq::context_t ctx;
-   zmq::socket_t sock(ctx, zmq::socket_type::push);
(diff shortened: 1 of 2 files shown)


sigiesec:
Hm, I don't think it is a good idea to add this. The README file is part of the repository, so it is implicitly clear that it refers to the current repository state. This may be wrong, but in that case it is clear it must be fixed. Adding a reference to a particular version only adds another potential place which may go out of sync. In particular, when you modify the example in the future, you would need to check which older version it is still working with.

gummif:
Good point. How about only specifying that the example requires C++11, since 98 is still technically supported?

sigiesec:
Yes, specifying that it requires C++11 is a good idea!

gummif:
Well I just realized that since it uses string_view it actually requires C++17...

gummif:
This PR ready now, using str_buffer instead of string_view.

sigiesec:
Great, thanks!

PR #339, opened 2019-08-15
Problem: C++14 features not detected using MSVC

gummif:
Solution: Detect C++14 via C++17


sigiesec:
Hm, for some reason I cannot merge this. Could you rebase it to the current maater?

PR #340, opened 2019-08-30
Fix ASCII character detection in conversion to std::string

boldorider4:
the 'Space' ASCII character should also be recognized as such.


sigiesec:
Thanks! If you find time it would be great if you could also add a test case for that!

Issue #335, opened 2019-07-15
'is_trivially_copyable' is not a member of std


bdsmith53:
Hi,

I am trying to use the chrono and thread libraries along with this binding for ZeroMQ. 

These libraries require the flag: -std=c++11 for the ISO C++ 2011 standard and support.

However, once I add this flag I get the following:

`In file included from thread_1.cpp:1:0:
/usr/local/include/zmq.hpp:334:19: error: ‘is_trivially_copyable’ is not a member of ‘std’
                && std::is_trivially_copyable<detail::range_value_t<Range>>::value
                   ^
/usr/local/include/zmq.hpp:332:39: error: parse error in template argument list
              typename = typename std::enable_if<
                                       ^
/usr/local/include/zmq.hpp:334:75: error: expected ‘>’ before ‘::’ token
                && std::is_trivially_copyable<detail::range_value_t<Range>>::value
                                                                           ^
/usr/local/include/zmq.hpp:335:60: error: ‘type’ in namespace ‘::’ does not name a type
                && !std::is_same<Range, message_t>::value>::type>
                                                            ^
/usr/local/include/zmq.hpp:951:7: error: ‘is_trivially_copyable’ is not a member of ‘std’
       std::is_trivially_copyable<T>::value && std::is_standard_layout<T>::value;
       ^
/usr/local/include/zmq.hpp:951:35: error: expected primary-expression before ‘>’ token
       std::is_trivially_copyable<T>::value && std::is_standard_layout<T>::value;
                                   ^
/usr/local/include/zmq.hpp:951:36: error: ‘::value’ has not been declared
       std::is_trivially_copyable<T>::value && std::is_standard_layout<T>::value;
`

hurdad:
I had same issue. looks like that function is not part of C++11, but newer :(

sigiesec:
std::is_trivially_copyable is defined by C++11, see e.g. https://en.cppreference.com/w/cpp/types/is_trivially_copyable. Maybe your toolset doesn't implement C++11 fully. Please add information on the platform (OS, compiler, standard library implementation, and their versions) you are using.

hurdad:
CentOS 7 GCC 4.8.5

bdsmith53:
Hi,

I solved this by using version 4.3.0 of cppzmq and the most recent version of libzmq (4.3.2).
GCC/G++ 4.9.2
Running on a Raspberry Pi: Raspbian GNU/Linux 8 (jessie)

[...]

PR #341, opened 2019-08-30
Problem: C++11 partially supported on gcc 4.8

gummif:
Solution: Use intrinsic instead of std::is_trivially_copyable for gcc versions older than 5.


gummif:
This is a fix for issue #335

sigiesec:
Great! Thanks in particular for adding CI with an older gcc.

PR #342, opened 2019-09-02
Move header files to an include directory

gummif:
It is standard practice to have exported headers in a directory called `include`. I would have liked to move the code to `include/cppzmq` but that would be a breaking change forcing code to do `#include <cppzmq/zmq.hpp>` (maybe for the next major release?).


bluca:
please don't, this will break a lot of scripts and packages

gummif:
@bluca Yes it will for non-cmake consumers. Would it make you more comfortable if this change were made in a major release (v5.0)?

bluca:
Sorry, but I don't feel like a cosmetic change on the repository layout is a good enough reason to break compatibility for all distros, regardless of versioning

sigiesec:
While I don't particularly like the current layout either, I agree with @bluca here. We only have two header files, so the situation on our side is not bad enough to warrant all the troubles a layout change might cause.

gummif:
I understand the concerns and this PR can be closed. Although I disagree that code layout change are unacceptable at the same time breaking code changes are introduced with a major version. I guess that's a conflict between long term backwards compatibility and using new features and best practices.

bluca:
Breaking backward compatibility can happen and the process is defined in C4 - but given the huge inconvenience, it should balanced by a very compelling requirement

PR #343, opened 2019-09-03
Problem: Sending string literals is awkward

gummif:
Solution: A function str_buffer specifically for
creating buffers for null terminated string literals.

Before
```C++
s.send(zmq::buffer(std::string_view{"ping"})); // C++17
auto s = "ping";
s.send(zmq::buffer(s, std::strlen(s))); // C++11
```
After
```C++
s.send(zmq::str_buffer("ping"));
```

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -1110,6 +1110,19 @@ const_buffer buffer(std::basic_string_view<T, Traits> data, size_t n_bytes) noex
 }
 #endif
 
+// Buffer for a string literal (null terminated)
+// where the buffer size excludes the terminating character.
+// Equivalent to zmq::buffer(std::string_view("...")).
+template<class Char, size_t N>
+const_buffer str_buffer(const Char (&data)[N]) noexcept
+{
+    static_assert(detail::is_pod_like<Char>::value, "Char must be POD");
+    static_assert(N > 0, "N > 0");
+    assert(data[N - 1] == Char{0});
+    return const_buffer(N == 1 ? nullptr : static_cast<const Char*>(data), 
+                        (N - 1) * sizeof(Char));
+}
+
 #endif // ZMQ_CPP11
 
 namespace detail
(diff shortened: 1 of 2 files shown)


zmq.hpp
```diff
@@ -1110,6 +1110,19 @@ const_buffer buffer(std::basic_string_view<T, Traits> data, size_t n_bytes) noex
 }
 #endif
 
+// Buffer for a string literal (null terminated)
+// where the buffer size excludes the terminating character.
+// Equivalent to zmq::buffer(std::string_view("...")).
+template<class Char, size_t N>
+const_buffer str_buffer(const Char (&data)[N]) noexcept
```
sigiesec: I am not 100% sure, but couldn't this be constexpr? (Maybe the assert conflicts with that, but then it could be constexpr at least with NDEBUG defined). Maybe the assertion could be avoided entirely if this were a user-defined literal operator instead?

zmq.hpp
```diff
@@ -1110,6 +1110,19 @@ const_buffer buffer(std::basic_string_view<T, Traits> data, size_t n_bytes) noex
...
+{
+    static_assert(detail::is_pod_like<Char>::value, "Char must be POD");
+    static_assert(N > 0, "N > 0");
```
sigiesec: An array cannot be 0-length, so this assertion is redundant.

zmq.hpp
```diff
@@ -1110,6 +1110,19 @@ const_buffer buffer(std::basic_string_view<T, Traits> data, size_t n_bytes) noex
...
+    assert(data[N - 1] == Char{0});
+    return const_buffer(N == 1 ? nullptr : static_cast<const Char*>(data),
```
sigiesec: Why handle the case of N==1 specially? Obviously, data[0] is valid in that case too, so returning nullptr in this case only opens up for additional pitfalls.

gummif: The specification for the buffer functions required 0 sized buffers to have nullptr. I will remove this check since it is not strictly required for this function.

gummif:
I added user-defined literals to make sending strings even more readable
```c++
s.send("ping"_zbuf);
```

blue7676:
I.     Don't.  Know.    What    I did!!!  Keep telling you that!!

sigiesec:
Thanks!

PR #349, opened 2019-09-17
Problem: Build error with clang

gummif:
Solution: Some compilers incorrectly define `__GNUC__`, added workaround for clang and icc.


blue7676:
Sorry guys but you know why much more than me am just going touninstall it
because I need to go back and learn some more before I even try this thanks
anyways

PR #355, opened 2019-10-12
C++17 library fixes

phlptp:
Ran into a couple issues using cppzmq in our test suite.  The specific cases that gave the most trouble were Clang 5 and 3.6.    The issue was not with the compiler itself, but in the test they are linking with a slightly older version of libstdc++,  which meant in the case of using C++17,  the headers for <optional> and <string_view> were not actually available even though ZMQ_CPP17 was enabled.   

This required some additional checks to ensure that the headers were available and a means to deactivate the particular feature.  

The second on Clang 3.6 was linking with a libstdc++ from GCC 4.9 which doesn't have the is_trivially_copyable based on the previous condition.    This required adding a condition specific to libstdc++ and a version date.    This is not perfect as there are some older versions of libstdc++ from GCC 5 that should work, but I didn't want to make the condition too complicated, since it would be a rather unusual circumstance.

The final fix in this PR is for zmq_addon.    Using <zmq.hpp> in certain compilers led it to not be found if zmq.hpp was not directly on the include search path.  Using `#include "zmq.hpp"  also lets compilers search in the current directory.   which is required for use in our application.  

So In summary this PR includes 3 fixes
- check C++17 headers directly for inclusion of <optional> and <string_view>
- check condition around `is_trivially_copyable` for cases where libstdc++ is used with Clang or ICC
- change zmq_addon.hpp to include zmq.hpp by quotes to allow searching current directory

It had been about a year since I last merged cppzmq with our local copy and like a lot of the recent changes even though they did lead to a lot of deprecated warnings that needed fixing in our code.

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -87,7 +87,17 @@
 #include <memory>
 #endif
 #ifdef ZMQ_CPP17
+#ifdef __has_include
+#if __has_include(<optional>)
 #include <optional>
+#define ZMQ_HAS_OPTIONAL 1
+#endif
+#if __has_include(<string_view>)
+#include <string_view>
+#define ZMQ_HAS_STRING_VIEW 1
+#endif
+#endif
+
 #endif
 
 /*  Version macros for compile-time API version detection                     */
@@ -127,12 +137,17 @@
 #if defined(ZMQ_CPP11) && !defined(__llvm__) && !defined(__INTEL_COMPILER) \
     && defined(__GNUC__) && __GNUC__ < 5
 #define ZMQ_CPP11_PARTIAL
+#elif defined(__GLIBCXX__) && __GLIBCXX__ < 20160805
+//the date here is the last date of gcc 4.9.4, which
+// effectively means libstdc++ from gcc 5.5 and higher won't trigger this branch
+#define ZMQ_CPP11_PARTIAL
 #endif
 
 #ifdef ZMQ_CPP11
 #ifdef ZMQ_CPP11_PARTIAL
 #define ZMQ_IS_TRIVIALLY_COPYABLE(T) __has_trivial_copy(T)
 #else
+#include <type_traits>
 #define ZMQ_IS_TRIVIALLY_COPYABLE(T) std::is_trivially_copyable<T>::value
 #endif
 #endif
@@ -700,7 +715,7 @@ struct recv_buffer_size
 namespace detail
 {
 
-#ifdef ZMQ_CPP17
+#if defined(ZMQ_HAS_OPTIONAL) && (ZMQ_HAS_OPTIONAL > 0)
 using send_result_t = std::optional<size_t>;
 using recv_result_t = std::optional<size_t>;
 using recv_buffer_result_t = std::optional<recv_buffer_size>;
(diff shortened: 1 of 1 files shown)


zmq.hpp
```diff
@@ -700,7 +714,7 @@ struct recv_buffer_size
 namespace detail
 {
 
-#ifdef ZMQ_CPP17
+#if ZMQ_HAS_OPTIONAL > 0
```
gummif: should this not be #ifdef ZMQ_HAS_OPTIONAL? (same below)

zmq_addon.hpp
```diff
@@ -24,7 +24,7 @@
 #ifndef __ZMQ_ADDON_HPP_INCLUDED__
 #define __ZMQ_ADDON_HPP_INCLUDED__
 
-#include <zmq.hpp>
+#include "zmq.hpp"
```
gummif: I'm not sure what I think of this. Is this not just a matter of adding the proper include path to the compiler?

zmq.hpp
```diff
@@ -87,7 +87,16 @@
 #include <memory>
 #endif
 #ifdef ZMQ_CPP17
+#ifdef __has_include
+#if __has_include(<optional>)
 #include <optional>
+#define ZMQ_HAS_OPTIONAL 1
+#endif
+#if __has_include(<string_view>)
+#define ZMQ_HAS_STRING_VIEW 1
```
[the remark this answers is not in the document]
phlptp: I was wondering that too,  My understanding of the standard is that there is a constructor for std::string which uses a string_view in the same way as it is being used here,  which means including std::string should be sufficient for this purpose.    I haven't used C++17 mode much and probably won't until next year when we start to switch to C++17,   So I don't have a strong opinion on that, as best as I can tell it is correct either way.

zmq.hpp
phlptp: I believe all preprocessors I know of at least will interpret this correctly, but I suppose
# if defined(ZMQ_HAS_OPTIONAL) && ZMQ_HAS_OPTIONAL>0 would be more correct.  I thought having a value associated with it would allow someone to turn it off even in C++17 mode,  where as just having the test check for definition would not allow that usage.

zmq_addon.hpp
phlptp: Yes, but when cppzmq is included in a code base as part of a set of 3rd party headers, all called like "libname/header.h"  having to add an extra include to a bunch of projects and propagate that through, it gets rather annoying to have to treat cppzmq as a special in some way.  So I have to make the change when I use it, so it is consistent with the other ThirdParty libraries we use.  

I am not going to be pedantic about it so if you want me change it back for the PR that is fine.

gummif: OK, I guess it's fine then since cppzmq will probably never move the headers into a cppzmq subfolder.

sigiesec: Probably using "" here is actually better, for the reason @phlptp mentioned, and also to ensure that the correct version of zmq.hpp is included. 

However, it would be good to have this in a separate commit, since it is not related to the other changes, and those wouldn't have touched zmq_addon.hpp otherwise.

zmq.hpp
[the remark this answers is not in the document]
sigiesec: Since you test `__has_include(<string_view>)`, it would be consistent to actually include `<string_view>`  in that case, even if it is not necessary for known compilers/library implementations.

sigiesec:
Thanks a lot for these improvements!

PR #356, opened 2019-10-16
Switch zmq_addon.hpp to include zmq.hpp using "zmq.hpp" instead

phlptp:
Switch zmq_addon.hpp to include zmq.hpp using "zmq.hpp" instead of <zmq.hpp>  This allows it to look in the current directory for the header file.  This enables the ability to have cppzmq as a subdirectory of an included folder vs needing to have the cppzmq on the include search path, and makes it less likely the wrong version would get included if more than one zmq.hpp header happened to be available.

This is a second PR based on conversation from #355


sigiesec:
Thanks a lot for this as well!

PR #358, opened 2019-10-26
Problem: Handling multipart messages is complex

gummif:
Solution: Add generic algorithms for sending and receiving multipart
messages.

A possible future extension to these functions is a `recv_multipart_n` where for writing to fixed size containers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,9 +41,8 @@ Supported platforms
 
 Examples
 ========
-This example requires at least C++11.
+These examples requires at least C++11.
 ```c++
-#include <string>
 #include <zmq.hpp>
 
 int main()
@@ -52,6 +51,36 @@ int main()
     zmq::socket_t sock(ctx, zmq::socket_type::push);
     sock.bind("inproc://test");
     sock.send(zmq::str_buffer("Hello, world"), zmq::send_flags::dontwait);
+}
+```
+This a more complex example where we send and receive multi-part messages.
+```c++
+#include <iostream>
+#include <zmq_addon.hpp>
+
+int main()
+{
+    zmq::context_t ctx;
+    zmq::socket_t sock1(ctx, zmq::socket_type::pair);
+    zmq::socket_t sock2(ctx, zmq::socket_type::pair);
+    sock1.bind("inproc://test");
+    sock2.connect("inproc://test");
+
+    std::array<zmq::const_buffer, 2> send_msgs = {
+        zmq::str_buffer("foo"),
+        zmq::str_buffer("bar!")
+    };
+    if (!zmq::send_multipart(sock1, send_msgs))
+        return 1;
+
+    std::vector<zmq::message_t> recv_msgs;
+    const auto ret = zmq::recv_multipart(
+        sock2, std::back_inserter(recv_msgs));
+    if (!ret)
+        return 1;
+    std::cout << "Got " << *ret
+              << " messages" << std::endl;
+    return 0;
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Supported platforms
 
 Examples
 ========
-These examples requires at least C++11.
+These examples require at least C++11.
 ```c++
 #include <zmq.hpp>
 
--- a/zmq_addon.hpp
+++ b/zmq_addon.hpp
@@ -96,8 +96,10 @@ template<class Range,
 detail::send_result_t send_multipart(socket_ref s, Range&& msgs,
                                        send_flags flags = send_flags::none)
 {
-    auto it = msgs.begin();
-    const auto end_it = msgs.end();
+    using std::begin;
+    using std::end;
+    auto it = begin(msgs);
+    const auto end_it = end(msgs);
     size_t msg_count = 0;
     while (it != end_it)
     {
@@ -106,7 +108,7 @@ detail::send_result_t send_multipart(socket_ref s, Range&& msgs,
         if (!s.send(*it, msg_flags))
         {
             // zmq ensures atomic delivery of messages
-            assert(it == msgs.begin());
+            assert(it == begin(msgs));
             return {};
         }
         ++msg_count;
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -41,9 +41,8 @@ Supported platforms
 
 Examples
 ========
-This example requires at least C++11.
+These examples requires at least C++11.
```
sigiesec:
```suggestion
These examples require at least C++11.
```

zmq_addon.hpp
```diff
@@ -37,6 +37,78 @@
...
+#ifdef ZMQ_CPP11
+
+/*  Receive a multipart message.
+    
+    Writes the zmq::message_t objects to OutputIterator out.
+    The out iterator must handle an unspecified amount of write,
+    e.g. using std::back_inserter.
+    
+    Returns: the number of messages received or nullopt (on EAGAIN).
+    Throws: if recv throws.
```
sigiesec: `recv` can probably only throw on the first message part, so in that case, no parts will have been written to `out`.

However, this throws also when `*out++` throws (e.g. `std::bad_alloc`), in which case we are left with a have-received message, of which one part is lost. Maybe that's fine, but should be specified. I am not sure if it would be better to `recv` the remaining parts as well and drop them, to ensure a more consistent state.

zmq_addon.hpp
```diff
@@ -37,6 +37,78 @@
 
 namespace zmq
 {
+
+#ifdef ZMQ_CPP11
+
+/*  Receive a multipart message.
+    
+    Writes the zmq::message_t objects to OutputIterator out.
+    The out iterator must handle an unspecified amount of write,
```
sigiesec:
```suggestion
    The out iterator must handle an unspecified number of writes,
```
See also comment below on throwing.

zmq_addon.hpp
```diff
@@ -37,6 +37,78 @@
...
+               detail::is_range<Range>::value
+               && (std::is_same<detail::range_value_t<Range>, message_t>::value
+               || detail::is_buffer<detail::range_value_t<Range>>::value)
+               >::type>
+detail::send_result_t send_multipart(socket_ref s, Range&& msgs,
+                                       send_flags flags = send_flags::none)
+{
+    auto it = msgs.begin();
+    auto last = msgs.end();
+    const size_t msg_count = static_cast<size_t>(std::distance(it, last));
```
sigiesec:
```suggestion
    const auto msg_count = static_cast<size_t>(std::distance(it, last));
```
since the type is already explicit in the initializer.

However, the value is not used right now (except for returning it), and for a non-random-access iterators, this is rather inefficient. We can also count while iterating, if the number of elements should really be returned (I am not sure if that is necessary: if it is relevant to the caller, they could also determine the size of the range themselves; however, it might make sense for consistency reasons).

zmq_addon.hpp
```diff
@@ -37,6 +37,78 @@
...
+             typename = typename std::enable_if<
+               detail::is_range<Range>::value
+               && (std::is_same<detail::range_value_t<Range>, message_t>::value
+               || detail::is_buffer<detail::range_value_t<Range>>::value)
+               >::type>
+detail::send_result_t send_multipart(socket_ref s, Range&& msgs,
+                                       send_flags flags = send_flags::none)
+{
+    auto it = msgs.begin();
+    auto last = msgs.end();
```
sigiesec:
```suggestion
    const auto last = msgs.end();
```

sigiesec: Also, please do not call this `last`, but rather `end`. `last` suggests that this points to the last *valid* element. It has a reason that it is called the `end`, and the terminology should be kept consistent.

zmq_addon.hpp
```diff
@@ -37,6 +37,78 @@
...
+}
+
+/*  Send a multipart message.
+    
+    The range must be a ForwardRange of zmq::message_t,
+    zmq::const_buffer or zmq::mutable_buffer.
+    The flags may be zmq::send_flags::sndmore if there are 
+    more message parts to be sent after the call to this function.
+    
+    Returns: the number of messages sent or nullopt (on EAGAIN).
```
sigiesec: The number of messages sent is always the size of the range. It should be made clear that the result cannot be that only a fragment of the multipart message has been sent.

zmq_addon.hpp
```diff
@@ -37,6 +37,78 @@
...
+    auto it = msgs.begin();
+    auto last = msgs.end();
+    const size_t msg_count = static_cast<size_t>(std::distance(it, last));
+    for (; it != last; ++it)
+    {
+        const auto mf = flags | (std::next(it) == last ? send_flags::none : send_flags::sndmore);
+        if (!s.send(*it, mf))
+        {
+            // zmq ensures atomic delivery of messages
+            assert(it == msgs.begin());
```
sigiesec: This result of `std::next` is essentially discarded, which may be inefficient, depending on the iterator implementation. If we determine the `msg_count` above, we can check the position via the index instead. An additional loop variable is then required.

If `msg_count` were dropped, then this might be acceptable, or this might be changed to something like:
```
for (; it != last;) {
   auto msg = std::move(*it);
   ++it;
   if (!s.send(msg, flags | (it == last ? send_flags::none : send_flags::sndmore)) {
      assert(it == std::next(msgs.begin()));
```

zmq_addon.hpp
gummif: Ah yes, good point. In that unlikely event I think the caller should handle it, probably by reconnecting the socket, or maybe terminate in the case of bad_alloc.

zmq_addon.hpp
gummif: Yes that's true.

zmq_addon.hpp
gummif: I am following the convention on cppreference.com but I agree the naming is unfortunate.

zmq_addon.hpp
gummif: Fixed to do only one increment inside the loop.

zmq_addon.hpp
sigiesec: Oh really? I am sorry then regarding the remark on consistency. Maybe I am mistaken about the use of 'last', and that's only my personal feeling.

gummif: That's ok, I don't have a strong opinion on that.

PR #366, opened 2019-11-09
Problem: Detail namespace used in API

gummif:
Solution: Directly use optional type

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -712,14 +712,16 @@ struct recv_buffer_size
     }
 };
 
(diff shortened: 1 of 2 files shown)


sigiesec:
Actually, I liked the fact that there are type aliases for the result types. I think this is more expressive. Why don't you just move the types from `zmq::detail` to `zmq`?

gummif:
Yes good point. Fixed.

sigiesec:
Great, thanks! Sorry for the delay.

PR #370, opened 2019-12-05
Problem: Missing recv multipart to fixed buffers

gummif:
Solution: Add recv_multipart_n function

### DIFF
--- a/zmq_addon.hpp
+++ b/zmq_addon.hpp
@@ -40,6 +40,41 @@ namespace zmq
 
 #ifdef ZMQ_CPP11
 
+namespace detail
+{
+template<bool CheckN, class OutputIt>
+recv_result_t recv_multipart_n(socket_ref s, OutputIt out, size_t n,
+                               recv_flags flags)
+{
+    size_t msg_count = 0;
+    message_t msg;
+    while (true)
+    {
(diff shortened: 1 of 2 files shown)


zmq_addon.hpp
```diff
@@ -40,27 +40,25 @@ namespace zmq
...
 {
     size_t msg_count = 0;
     message_t msg;
     while (true)
     {
+        #ifdef ZMQ_CPP17
+        if constexpr (CheckN)
+        #else
+        if (CheckN)
+        #endif
```
sigiesec: One minor comment on this: I think it is not worthwhile to conditionally use `constexpr` here.

While I really appreciate `if constexpr`, I think of it primarily as a means to simplify code, e.g. where previously template specializations or multiple SFINAE overloads needed to be used. It is necessary when the code in the branch is not valid when the condition is false. This is not the case here. The compiler is easily able to optimize the if away if the condition is false without `constexpr`. Here, due to the necessity of the `#ifdef`, the code is more complex with the `constexpr` than without, so I would rather leave it out.

gummif: Yes, good point.

sigiesec:
Great, thanks :) And thanks again for adding comprehensive tests along with the new function!

PR #371, opened 2019-12-05
Problem: Move assigned objects still alive

gummif:
Solution: Close context/socket if move assigned to

The problem is the following
```c++
void f(socket_t& a, socket_t& b)
{
  a = std::move(b);
}
```
Currently after calling this function `b` will be the socket that used to be `a`, which I think is surprising. After this PR `b` will be an "empty" socket.
Since this is only changing the value of a moved-from objects it should not break user code, unless they are doing something they shouldn't be doing.


sigiesec:
Oh, yes, that's really surprising behaviour. I am a bit curious though how you use `zmq::socket_t` in a way that it is re-assigned from another object?

PR #374, opened 2020-01-15
Problem: Invalid 32bit issues when using Conan

gummif:
Solution: Remove architechture checks from cmake version file

Since this is a header only library there is no need to check for 32 vs 64bits.

See similar problem: https://github.com/onqtam/doctest/pull/225


sigiesec:
It's not that obvious why this is necessary. Could you maybe extend the comment so that this can be understood easier, and the risk someone removes it accidentally again is reduced? I think this is quite a nice explanation: https://github.com/catchorg/Catch2/blob/master/CMakeLists.txt#L140

sigiesec:
Thanks!

PR #377, opened 2020-01-21
Problem: shutdown() missing for context_t

gummif:
Solution: Add shutdown(). This function is required
for clean termination of the zmq context in multi-threaded
applications where sockets are used in threads. In particular
if blocking operation are used and if sockets can be created
at any time.

### DIFF
--- a/tests/context.cpp
+++ b/tests/context.cpp
@@ -18,6 +18,26 @@ TEST_CASE("context create, close and destroy", "[context]")
     CHECK(NULL == (void *) context);
 }
 
+TEST_CASE("context shutdown", "[context]")
+{
+    zmq::context_t context;
+    context.shutdown();
+    CHECK(NULL != (void *) context);
+    context.close();
+    CHECK(NULL == (void *) context);
+}
+
+TEST_CASE("context shutdown often", "[context]")
+{
+    zmq::context_t context;
+    context.shutdown();
+    context.shutdown();
+    context.shutdown();
+    CHECK(NULL != (void *) context);
+    context.close();
+    CHECK(NULL == (void *) context);
+}
+
 #ifdef ZMQ_CPP11
 TEST_CASE("context swap", "[context]")
 {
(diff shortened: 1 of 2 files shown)


tests/context.cpp
```diff
@@ -18,6 +18,26 @@ TEST_CASE("context create, close and destroy", "[context]")
...
+    context.close();
+    CHECK(NULL == (void *) context);
+}
+
+TEST_CASE("context shutdown often", "[context]")
+{
+    zmq::context_t context;
+    context.shutdown();
+    context.shutdown();
+    context.shutdown();
```
sigiesec: Calling it three times is a bit confusing. To check that it can be called more than once, two calls were sufficient. To check if it can still be called when all sockets have been closed or so, a different test would be needed.

zmq.hpp
```diff
@@ -691,6 +692,19 @@ class context_t
         ptr = ZMQ_NULLPTR;
     }
 
+    // Shutdown context in preparation for termination (close()).
+    // Causes all blocking socket operations and any further
+    // operations to return with ETERM.
+    // Operation on sockets constructed after
+    // this call will however NOT return with ETERM.
```
sigiesec: Is that so? The wording on http://api.zeromq.org/4-3:zmq-ctx-shutdown is not clear about that, but this doesn't sound intended. However, I would leave this out here, and rather open an issue with libzmq to clarify and maybe fix that (I will try to give that a closer look myself, I found there is no test for this).

gummif: Yes that was a bit surprising. This code will block forever:
```c++
zmq::context_t context;
context.shutdown();
zmq::socket_t sock(context, zmq::socket_type::rep);
sock.connect("inproc://test");
zmq::message_t msg;
sock.recv(msg);
```
That is the reason for my other PR https://github.com/zeromq/cppzmq/pull/378. Because a thread that might open a new socket just after a shutdown will block forever on recv.

sigiesec: Thanks a lot for bringing this up and opening zeromq/libzmq#3792!

Could you remove the comment for now? You edited it such that it is correct in any case, but I think this is rather confusing.

If zeromq/libzmq#3792 were closed without fixing that, we can revisit that.

Similarly, I'd also defer #378 if that's ok for you. It somehow contradicts the light-weight nature of cppzmq to add an extra class just to work around a bug in libzmq. For most bugs, this is not even possible, just for this particular case it might be.

sigiesec:
Thanks for the PR, the discussion and bringing the libzmq issue up!

PR #378, opened 2020-01-21
Problem: Terminating ctx safely is hard

gummif:
Solution: A shudown_guard running in a thread
that calls shutdown() on a context until
notified to stop. This class makes terminating
a context in multi-threaded code safe and easy.

### DIFF
--- a/zmq_addon.hpp
+++ b/zmq_addon.hpp
@@ -31,7 +31,10 @@
 #include <sstream>
 #include <stdexcept>
 #ifdef ZMQ_CPP11
+#include <condition_variable>
 #include <functional>
+#include <mutex>
+#include <thread>
 #include <unordered_map>
 #endif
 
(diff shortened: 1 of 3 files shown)


gummif:
Looks like I have to make the tests a bit more robust with respect to race conditions.

gummif:
The background on this is that in our code we were sometimes experiencing crashes and hangs. When the code was ran with ThreadSanitizer it pointed to data race in close() in context_t. We added a timed mutex around the context to synchronize, and that fixed most of the issues but there were sometimes still some data races reported. This appears to fix these issues.

Not related to this PR, I'm still however getting a data race between thread connecting to/receiving from a socket and a "ZMQbg/IO/1" thread. Need to investigate that further.

zmq_addon.hpp
```diff
@@ -572,6 +575,62 @@ class active_poller_t
...
+                            std::chrono::microseconds iv = std::chrono::microseconds(5000))
+    {
+        thrd = std::thread([&, iv]{
+            while (true)
+            {
+                ctx.shutdown();
+                std::unique_lock<std::mutex> lock{mtx};
+                if (cv.wait_for(lock, iv, [this]{ return do_stop; }))
+                    return; // stop requested
+            }
```
sigiesec: I somehow don't get the point on the use case for this. Why does this need to be a loop which might call shutdown multiple times? When would a single call to shutdown not be sufficient?

gummif: This might not be required if 1 socket is created before shutdown, see https://github.com/zeromq/libzmq/issues/3792. A better workaround/hack is maybe to create a temporary socket in context_t::shutdown before calling zmq_ctx_shutdown?

gummif:
Closing this since shutdown has been fixed in libzmq https://github.com/zeromq/libzmq/pull/3794

PR #380, opened 2020-01-26
Reformat all files with clang-format

sigiesec:
We have a .clang-format file but over the past 1.5 years several deviations from the defined formatting have evolved. This reformats all files with clang-format to reestablish consistent formatting.


sigiesec:
@zeromq/core Can someone merge this, please?

PR #383, opened 2020-02-04
Problem: Context lacks typesafe options

gummif:
Solution: Define an enum class for the context options

I'm not sure what to do about the options that have a value of -1, with this change trying to get them results in an exception with the last set errno.

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -643,7 +643,7 @@ enum class ctxopt
     thread_affinity_cpu_remove = ZMQ_THREAD_AFFINITY_CPU_REMOVE,
     thread_name_prefix = ZMQ_THREAD_NAME_PREFIX,
     max_msgsz = ZMQ_MAX_MSGSZ,
-#ifdef ZMQ_BUILD_DRAFT_API
+#ifdef ZMQ_ZERO_COPY_RECV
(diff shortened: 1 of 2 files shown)


zmq.hpp
```diff
@@ -630,6 +632,27 @@ inline void swap(message_t &a, message_t &b) ZMQ_NOTHROW
...
+{
+    blocky = ZMQ_BLOCKY,
+    io_threads = ZMQ_IO_THREADS,
+    thread_sched_policy = ZMQ_THREAD_SCHED_POLICY,
+    thread_priority = ZMQ_THREAD_PRIORITY,
+    thread_affinity_cpu_add = ZMQ_THREAD_AFFINITY_CPU_ADD,
+    thread_affinity_cpu_remove = ZMQ_THREAD_AFFINITY_CPU_REMOVE,
+    thread_name_prefix = ZMQ_THREAD_NAME_PREFIX,
+    max_msgsz = ZMQ_MAX_MSGSZ,
+#ifdef ZMQ_BUILD_DRAFT_API
```
sigiesec: Hm, this should probably say 
```suggestion
#ifdef ZMQ_ZERO_COPY_RECV
```
since not every (older) version of libzmq has it, even when `ZMQ_BUILD_DRAFT_API` is defined.

Issue #385, opened 2020-02-19
"Context lacks typesafe options" causes build error


lassoan:
This [recent commit](https://github.com/zeromq/cppzmq/commit/324b11f23951127423f66839bdc8cad53ff76c8f) broke build of our project. Everything worked well before, but we get these build errors now (when building [this file](https://github.com/Slicer/SlicerJupyter/blob/master/JupyterKernel/xSlicerServer.cxx)):

```
C:\D\SlicerJupyter_R\cppzmq\zmq.hpp(642): error C2065: 'ZMQ_THREAD_AFFINITY_CPU_ADD': undeclared identifier (compiling
source file C:\D\SlicerJupyter\JupyterKernel\xSlicerServer.cxx) [C:\D\SlicerJupyter_R\inner-build\JupyterKernel\qSlicer
JupyterKernelModule.vcxproj] [C:\D\SlicerJupyter_R\inner.vcxproj]
C:\D\SlicerJupyter_R\cppzmq\zmq.hpp(643): error C2065: 'ZMQ_THREAD_AFFINITY_CPU_REMOVE': undeclared identifier (compili
ng source file C:\D\SlicerJupyter\JupyterKernel\xSlicerServer.cxx) [C:\D\SlicerJupyter_R\inner-build\JupyterKernel\qSli
cerJupyterKernelModule.vcxproj] [C:\D\SlicerJupyter_R\inner.vcxproj]
C:\D\SlicerJupyter_R\cppzmq\zmq.hpp(644): error C2065: 'ZMQ_THREAD_NAME_PREFIX': undeclared identifier (compiling sourc
e file C:\D\SlicerJupyter\JupyterKernel\xSlicerServer.cxx) [C:\D\SlicerJupyter_R\inner-build\JupyterKernel\qSlicerJupyt
erKernelModule.vcxproj] [C:\D\SlicerJupyter_R\inner.vcxproj]
C:\D\SlicerJupyter_R\cppzmq\zmq.hpp(652): error C2065: 'ZMQ_MSG_T_SIZE': undeclared identifier (compiling source file C
:\D\SlicerJupyter\JupyterKernel\xSlicerServer.cxx) [C:\D\SlicerJupyter_R\inner-build\JupyterKernel\qSlicerJupyterKernel
Module.vcxproj] [C:\D\SlicerJupyter_R\inner.vcxproj]
[...]

PR #386, opened 2020-02-19
Problem: Macros not defined in older versions 

gummif:
Solution: Guard usage of libzmq macros

Fixes https://github.com/zeromq/cppzmq/issues/385


sigiesec:
Thanks a lot! Also for adding CI, so we will not overlook a similar problem in the future!

PR #388, opened 2020-02-29
Problem: cppzmq lacks CZMQ compatible 'encode/decode' and 'actor'

brettviren:
This adds methods `encode()` and `decode()` to `multipart_t` which should be compatible with CZMQ's equivalent methods.  In developing the methods I copied CZMQ's code but admit I have not yet actually tried a live cross-code test.  There is a test added in cppzmq so the codec internally round-trips okay.

This PR also adds a new file `zmq_actor.hpp` which is an attempt to make a CZMQ-like actor pattern.  It attempts to copy the shutdown semantics directly from CZMQ.  It uses `std::thread` and has a `.join()` at shutdown which I can't find an equivalent for in CZMQ.  A test is added.  

Hopefully it's okay to bundle these two problems/solutions in one PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 
 # Build directory
 *build/
+*~
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,7 @@ message(STATUS "cppzmq v${cppzmq_VERSION}")
 set(CPPZMQ_HEADERS
     zmq.hpp
     zmq_addon.hpp
+    zmq_actor.hpp
 )
 
 foreach (target cppzmq cppzmq-static)
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -29,6 +29,7 @@ add_executable(
     send_multipart.cpp
     monitor.cpp
     utilities.cpp
+    actor.cpp
 )
 
 add_dependencies(unit_tests catch)
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -17,6 +17,7 @@ find_package(Threads)
 
 add_executable(
     unit_tests
+    actor.cpp
     buffer.cpp
     message.cpp
     context.cpp
@@ -29,7 +30,6 @@ add_executable(
     send_multipart.cpp
     monitor.cpp
     utilities.cpp
-    actor.cpp
 )
 
 add_dependencies(unit_tests catch)
--- a/tests/actor.cpp
+++ b/tests/actor.cpp
@@ -28,12 +28,11 @@ TEST_CASE("actor external and internal termination", "[actor]")
         zmq::actor_t actor(context, myactor, "hello world", false);
         actor.pipe().send(zmq::message_t{}, zmq::send_flags::none);
     }
+}
 
-// }
-
-// TEST_CASE("actor internal termination", "[actor]")
-// {
-//     zmq::context_t context;
+TEST_CASE("actor internal termination", "[actor]")
+{
+    zmq::context_t context;
     {
         zmq::actor_t actor(context, myactor, "fast exit", true);
         actor.pipe().send(zmq::message_t{}, zmq::send_flags::none);
--- a/tests/actor.cpp
+++ b/tests/actor.cpp
@@ -21,7 +21,7 @@ void myactor(zmq::socket_t& pipe, std::string greeting, bool fast_exit)
 }
 
 
-TEST_CASE("actor external termination", "[actor]")
+TEST_CASE("actor external and internal termination", "[actor]")
 {
     zmq::context_t context;
     {
@@ -29,11 +29,11 @@ TEST_CASE("actor external termination", "[actor]")
         actor.pipe().send(zmq::message_t{}, zmq::send_flags::none);
     }
 
-}
+// }
 
-TEST_CASE("actor internal termination", "[actor]")
-{
-    zmq::context_t context;
+// TEST_CASE("actor internal termination", "[actor]")
+// {
+//     zmq::context_t context;
     {
         zmq::actor_t actor(context, myactor, "fast exit", true);
         actor.pipe().send(zmq::message_t{}, zmq::send_flags::none);
--- a/tests/actor.cpp
+++ b/tests/actor.cpp
@@ -1,9 +1,8 @@
 #include <catch.hpp>
-#include <zmq_actor.hpp>
 
+#include <zmq_actor.hpp>
 #ifdef ZMQ_CPP11
 #include <future>
-#endif
 
 #if (__cplusplus >= 201703L)
 static_assert(std::is_nothrow_swappable<zmq::socket_t>::value,
@@ -40,3 +39,5 @@ TEST_CASE("actor internal termination", "[actor]")
         actor.pipe().send(zmq::message_t{}, zmq::send_flags::none);
     }
 }
+
+#endif  // ZMQ_CPP11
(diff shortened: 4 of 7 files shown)


brettviren:
From the travis tests, it looks like I need to add some protection for non-C++11 compilers.  

I'll work on that and update this PR.

brettviren:
I fixed the code problems but I'm not sure what is happening with this last test.

The last test initiates a shutdown by the main sending a signal through the actor pipe to the actor which is waiting in a `recv()`.  The `recv()` returns and the actor function exits.  This works locally but apparently hangs on travis.

bluca:
This project is licensed under MIT, while CZMQ is MPL2. Copying code changes the license of the whole file, unless you can get permission from the original author, in writing, to relicense those parts. Else, it needs to be reimplemented from scratch without looking at the original.

brettviren:
@bluca hmm, a difficult point.  

The proposed encode/decode code isn't a literal copy of course as CZMQ code is in terms of `zframe_t` and `zmsg_t` while cppzmq is in terms of `message_t` and `multipart_t`.  The few lines of actual bit packing are essentially the same as what is in CZMQ but it's hard to implement such a simple thing without those lines looking the same.

Do you have a suggestion for how I might provide compatible codec without violating license? 

I could close my eyes to the exact variable names and code forms in CZMQ and rewrite encode()/decode() if that would be clean enough.

brettviren:
BTW, the last check passed and I suspect that it did so because I form an inproc:// address using pointer values hoping for uniqueness.  The last commit runs the two tests in serial and passes so I suspect the failure in the previous commit was due to the address not actually being unique.  So, despite passing travis-ci, this last commit has a bug.  I must find a way to assure truly unique inproc:// address forming.

bluca:
If you make it different enough, that's good - the on-the-wire protocol is obviously fine to reimplement. I just want to make sure there's no copy/paste.

brettviren:
I am out of ideas why the travis-ci actor test fails.  The last change from pass to fail only removed a `cout` debug.  In all cases, the test passes locally.  

Maybe there is some race condition that only shows up in travis-ci?  

I've had these kind of "mysterious" failures on travis-ci in other projects.  Being able to reproduce them locally would give a huge benefit.

brettviren:
Well, simply changing the order of tests passes travis-ci.  If there is some race condition, this change may be merely hiding it.  Or, it may be legitimately working around whatever confuses travis-ci.

At this point I guess this PR is ready for merging (on my end).  If others have suggestions to resolve the uncertainty I'll follow up.

sigiesec:
Hi... sorry for not responding earlier. First of all, thanks a lot for your contribution.

I didn't look into the details yet, but let me share a general concern. As laid out in the README, cppzmq aims to be a thin wrapper around the C API of libzmq. It is not a high-level API like czmq which provides it's own C++ binding. So I have the impression that these additions are outside the scope of cppzmq itself.

Adding more header files to cppzmq itself is something I am also quite hesitant about, as it makes integration more complex.

I hope you understand these concerns. I will give the PR a closer look and try to find a good solution for dealing with that.

brettviren:
Hi @sigiesec thanks for giving the PR attention.

I understand your concern and that you need to make a judgement call.  Let me give a tiny bit of "lobbying".

First, please take the addition of `multipart_t:encode()/decode()` as orthogonal to the addition of `zmq_actor.hpp`.  I really should have PR'ed them individually.  If only one makes your cut I'll update the PR to remove the other.

The `encode()/decode()` is maybe more important to accept as it helps `multipart_t` be used with SERVER/CLIENT and in a way that increases ZeroMQ ecosystem compatibility (ie, same packing as CZMQ).  I see these two methods being on par with other "helper" methods in `multipart_t` that strictly could be implemented outside the class.

The lack of `zmq_actor.hpp` is something that has kept me from using cppzmq in past projects.  My current one is important enough and I want to use cppzmq that I reach the threshold to write it instead of using CPPZMQ.  I can see how you might see it as the first step of re-implementing all of CZMQ inside cppzmq, but that's at least not my goal. I see it on par with the current `zpp_addon.hpp` as that also could be provided by application-specific code instead of distributing with cppzmq.  It's on a spectrum of course, which is why I think it's fair for you to pass some judgement.  But, I don't understand the concern about how a 3rd header increases difficulty of integration.  Like `zmq_addon.hpp`, `zmq_actor.hpp` is a "take it or leave it" header if a user copies the `.hpp` files.  If the user does a full install then it produces effectively no more difficulty.  But, I could be missing some aspect here.

Looking forward to your decision!

kurdybacha:
It should be two separate PR @brettviren for easier review and and probably faster merge. Contribution guidelines (C4) say: `A patch SHOULD be a minimal and accurate answer to exactly one identified and agreed problem.` to help with smother merging your code really.

I think new actor addition is a useful piece of code and as long @bluca and @sigiesec are happy it is +1 from me.

brettviren:
Hi @kurdybacha. It seems at this point the issue is about the validity of the two problems and not yet so much the solutions.  I'd like to get the results of @sigiesec judgement on these before taking further action

- [ ] Problem: multipart_t can not be used with SERVER/CLIENT

- [ ] Problem: cppzmq lacks an actor pattern

brettviren:
oops, accidentally clicked close!

brettviren:
@sigiesec any thoughts yet?  I need to soon make some decisions about my own work that depend your judgement call.  I'm okay with which ever way it swings so don't worry about hurt feelings.  If one or both problems are not accepted I'll move them into my own project.  If one or both are accepted we can then move on to critiquing the solution(s) including me splitting the PR into two if that's needed.  Thanks!

sigiesec:
You're right, that it doesn't make a huge difference if there are two or three headers. `zmq_addon.hpp` already deviates to some degree from the design goals I describe. My concern is that this is one more step that opens up the scope of cppzmq, which may make it harder to maintain. However, maybe this is too speculative. If that really happens, and we feel that it would better be split up into separate libraries/repositories, we can do so in the future, though it will be a  breaking change then. At some point in the future, we might drop support for pre-C++11 users, and this will also be a breaking change. It may make sense to consider all such matters at once, in order not to make multiple subsequent breaking changes in a short time.

So I think that both aspects can be added to cppzmq at this point. Thanks again for contributing! I hope that my hesitance doesn't give you a bad feeling, and doesn't prevent you from future contributions. If you have any concerns in that direction, please let me know.

Let me make clear that I am not the one making final decisions here. We aim to follow the C4 process, but it doesn't make a clear statement on the scope question AFAIU.

That being said, the C4 process does encourage to make contributions separately for separable aspects. Each commit should transition from a green state to a green state though. Also, the commit history should remain readable. Therefore, I'd ask you to rewrite the changes to two or three separate commits: One dealing with zmq::multipart_t, one dealing with zmq_actor, and, if that should be included, the change to .gitignore. I will look at the details then, and maybe comment if necessary. Splitting into separate PRs isn't necessary, but, as @kurdybacha wrote, might speed up finalizing either aspect.

Are you ok with that?

brettviren:
Thanks, @sigiesec, I am very okay with this.  I will close this PR and open new ones each focused on one problem as @kurdybacha suggested and I'll squash commits to get atomic "green-to-green".

And, also no, there are no bad feelings at all.  I do understand your concern about increasing the scope.  I worry about this aspect when I sit in your role on other projects.

BTW, I personally like the idea of looking toward a future break point which leaves behind pre-C++11.  When that happens I hope the surviving `send()` and `recv()` can gain a default `*_flags::none` argument.

Thanks again for your and others' attention on this.

brettviren:
For reference, this is now split into #391 and #392.  

I omitted the little change to the `.gitignore` file.

PR #389, opened 2020-03-05
Wraping calls to find_package

SylvainCorlay:
Wrapping calls to `find_package` allows building cppzmq without doing `find_package(ZeroMQ)` if the libzmq target is already defined.

This occurs when building a project which vendors all of its dependencies with `add_subdirectory`, resulting in targets being defined and find_package to fail. 

Note: I agree that vendoring dependencies in generally bad practice but this is a requirement in our context as we are building a PyPI wheel.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@ include (DetectCPPZMQVersion)
 
 project(cppzmq VERSION ${DETECTED_CPPZMQ_VERSION})
 
-if (NOT TARGET libzmq)
+if (NOT TARGET libzmq AND NOT TARGET libzmq-static)
   find_package(ZeroMQ QUIET)
 
   # libzmq autotools install: fallback to pkg-config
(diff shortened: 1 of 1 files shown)


sigiesec:
Looks ok to me. I wonder if there is a generic solution for this in CMake? I guess this may be necessary for any library in such circumstances.

SylvainCorlay:
> I wonder if there is a generic solution for this in CMake?

Not to my knowledge. I have looked and found this pattern in a bunch of places...

SylvainCorlay:
Hey, is there a plan for a release soon that would include that change? I see that the master branch already indicates 4.7.

SylvainCorlay:
@sigiesec do you guys have plans for a release including these changes in the near future?

Also, I wanted to say kudos to the whole team working on cppzmq. Clear deprecation warnings, clean cmake... The project has come a long way in the past year.

Issue #365, opened 2019-11-09
Cannot use send_multipart as shown in example or test doc with c++11.


randallpittman:
OS: CentOs 7
libzmq: 5.2.3 (git master built from source)
Cloned and built latest source (a34d2a3).

Made this simple program:
```cpp
#include <zmq_addon.hpp>

int main() {
  zmq::context_t context(1);
  zmq::socket_t input(context, ZMQ_PAIR);
  input.connect("inproc://multipart.test"); 
  std::array<zmq::message_t, 2> imsgs = {zmq::message_t(3), zmq::message_t(4)};
  auto iret = zmq::send_multipart(input, imsgs);
  return 0;
}
```

Try to build with 
```bash
gcc -std=c++11 -lzmq test.cpp -o test.test
```

The build fails with the following error:
```
test.cpp:10:47: error: no matching function for call to \u2018send_multipart(zmq::socket_t&, std::array<zmq::message_t, 2ul>&)\u2019
   auto iret = zmq::send_multipart(input, imsgs);
                                               ^
test.cpp:10:47: note: candidate is:
In file included from test.cpp:1:0:
/usr/local/include/zmq_addon.hpp:96:23: note: template<class Range, class> zmq::detail::send_result_t zmq::send_multipart(zmq::socket_ref, Range&&, zmq::send_flags)
 detail::send_result_t send_multipart(socket_ref s, Range&& msgs,
                       ^
/usr/local/include/zmq_addon.hpp:96:23: note:   template argument deduction/substitution failed:

```

gummif:
What version of gcc is that? Can you tell me what the following return?
```c++
zmq::detail::is_range<std::array<zmq::message_t, 2>>::value
std::is_same<zmq::detail::range_value_t<std::array<zmq::message_t, 2>>, zmq::message_t>::value
```
Both should return true.

zhbbupt:
@gummif 
I get the same error. And I run 
zmq::detail::is_range<std::array<zmq::message_t, 2>>::value
std::is_same<zmq::detail::range_value_t<std::array<zmq::message_t, 2>>, zmq::message_t>::value

`zmq::detail::is_range<std::array<zmq::message_t, 2>>::value` return true
but `std::is_same<zmq::detail::range_value_t<std::array<zmq::message_t, 2>>, zmq::message_t>::value` get an error:
`zmq.hpp:212:78: error: call of overloaded 'begin(std::array<zmq::message_t, 2ul>&)' is ambiguous
[...]

PR #390, opened 2020-03-07
Problem: send_multipart fails on old gcc versions

gummif:
Solution: remove template type checks if there is only partial C++11 support


gummif:
Should fix https://github.com/zeromq/cppzmq/issues/365

sigiesec:
Ok, thanks! As long as there are no other overloads that might cause ambiguity, this is probably ok.

PR #391, opened 2020-03-07
Problem: lack conversion `message_t` and `multipart_t`

brettviren:
Solution: add encode/decode methods to `multipart_t` giving a codec compatible with the CZMQ equivalent.

This is part of what was initially #388.

### DIFF
--- a/tests/codec_multipart.cpp
+++ b/tests/codec_multipart.cpp
@@ -1,0 +1,168 @@
+#include <catch.hpp>
+#include <zmq_addon.hpp>
+
+#ifdef ZMQ_HAS_RVALUE_REFS
+
+TEST_CASE("multipart codec empty", "[codec_multipart]")
+{
+    using namespace zmq;
+
+    multipart_t mmsg;
+    message_t msg = mmsg.encode();
+    CHECK(msg.size() == 0);
+    
+    multipart_t mmsg2;
+    mmsg2.decode_append(msg);
+    CHECK(mmsg2.size() == 0);
+
+}
+
+TEST_CASE("multipart codec small", "[codec_multipart]")
+{
+    using namespace zmq;
+
+    multipart_t mmsg;
+    mmsg.addstr("Hello World");
+    message_t msg = mmsg.encode();
+    CHECK(msg.size() == 1 + 11); // small size packing
+
+    mmsg.addstr("Second frame");
+    msg = mmsg.encode();
+    CHECK(msg.size() == 1 + 11 + 1 + 12);
+
+    multipart_t mmsg2;
+    mmsg2.decode_append(msg);
+    CHECK(mmsg2.size() == 2);
+    std::string part0 = mmsg2[0].to_string();
+    CHECK(part0 == "Hello World");
+    CHECK(mmsg2[1].to_string() == "Second frame");
+}
+
+TEST_CASE("multipart codec big", "[codec_multipart]")
+{
+    using namespace zmq;
+
+    message_t big(495);         // large size packing
+    big.data<char>()[0] = 'X';
+
+    multipart_t mmsg;
+    mmsg.pushmem(big.data(), big.size());
+    message_t msg = mmsg.encode();
+    CHECK(msg.size() == 5 + 495);
+    CHECK(msg.data<unsigned char>()[0] == 0xFF);
+    CHECK(msg.data<unsigned char>()[5] == 'X');
+
+    CHECK(mmsg.size() == 1);
+    mmsg.decode_append(msg);
+    CHECK(mmsg.size() == 2);
+    CHECK(mmsg[0].data<char>()[0] == 'X');
+}
+
+TEST_CASE("multipart codec decode bad data overflow", "[codec_multipart]")
+{
+    using namespace zmq;
+
+    char bad_data[3] = {5, 'h', 'i'};
+    message_t wrong_size(bad_data, 3);
+    CHECK(wrong_size.size() == 3);
+    CHECK(wrong_size.data<char>()[0] == 5);
+    
+    CHECK_THROWS_AS(
+        multipart_t::decode(wrong_size),
+        std::out_of_range);
+}
+
+TEST_CASE("multipart codec decode bad data extra data", "[codec_multipart]")
+{
+    using namespace zmq;
+
+    char bad_data[3] = {1, 'h', 'i'};
+    message_t wrong_size(bad_data, 3);
+    CHECK(wrong_size.size() == 3);
+    CHECK(wrong_size.data<char>()[0] == 1);
+    
+    CHECK_THROWS_AS(
+        multipart_t::decode(wrong_size),
+        std::out_of_range);
+}
+
+
+// After exercising it, this test is disabled over concern of running
+// on hosts which lack enough free memory to allow the absurdly large
+// message part to be allocated.
+#if 0
+TEST_CASE("multipart codec encode too big", "[codec_multipart]")
+{
+    using namespace zmq;
+
+    const size_t too_big_size = 1L + 0xFFFFFFFF;
+    CHECK(too_big_size > 0xFFFFFFFF);
+    char* too_big_data = new char[too_big_size];
+    multipart_t mmsg(too_big_data, too_big_size);
+    delete [] too_big_data;
+
+    CHECK(mmsg.size() == 1);
+    CHECK(mmsg[0].size() > 0xFFFFFFFF);
+
+    CHECK_THROWS_AS(
+        mmsg.encode(),
+        std::range_error);
+}
+#endif
+
+TEST_CASE("multipart codec free function with vector", "[codec_multipart]")
+{
+    using namespace zmq;
+    std::vector<message_t> parts;
+    parts.emplace_back("Hello", 5);
+    parts.emplace_back("World",5);
+    auto msg = encode(parts);
+    CHECK(msg.size() == 1 + 5 + 1 + 5 );
+    CHECK(msg.data<unsigned char>()[0] == 5);
+    CHECK(msg.data<unsigned char>()[1] == 'H');
+    CHECK(msg.data<unsigned char>()[6] == 5);
+    CHECK(msg.data<unsigned char>()[7] == 'W');
+
+    std::vector<message_t> parts2;
+    decode(msg, std::back_inserter(parts2));
+    CHECK(parts.size() == 2);
+    CHECK(parts[0].size() == 5);
+    CHECK(parts[1].size() == 5);
+}
+
+TEST_CASE("multipart codec free function with multipart_t", "[codec_multipart]")
+{
+    using namespace zmq;
+    multipart_t mmsg;
+    mmsg.addstr("Hello");
+    mmsg.addstr("World");
+    auto msg = encode(mmsg);
+    CHECK(msg.size() == 1 + 5 + 1 + 5);
+    CHECK(msg.data<unsigned char>()[0] == 5);
+    CHECK(msg.data<unsigned char>()[1] == 'H');
+    CHECK(msg.data<unsigned char>()[6] == 5);
+    CHECK(msg.data<unsigned char>()[7] == 'W');
+
+    multipart_t mmsg2;
+    decode(msg, std::back_inserter(mmsg2));
+    CHECK(mmsg2.size() == 2);
+    CHECK(mmsg2[0].size() == 5);
+    CHECK(mmsg2[1].size() == 5);
+}
+
+TEST_CASE("multipart codec static method decode to multipart_t", "[codec_multipart]")
+{
+    using namespace zmq;
+    multipart_t mmsg;
+    mmsg.addstr("Hello");
+    mmsg.addstr("World");
+    auto msg = encode(mmsg);
+
+    auto mmsg2 = multipart_t::decode(msg);
+    CHECK(mmsg2.size() == 2);
+    CHECK(mmsg2[0].size() == 5);
+    CHECK(mmsg2[1].size() == 5);
+}
+
+
+#endif
--- a/tests/codec_multipart.cpp
+++ b/tests/codec_multipart.cpp
@@ -1,7 +1,7 @@
 #include <catch.hpp>
 #include <zmq_addon.hpp>
 
-#ifdef ZMQ_HAS_RVALUE_REFS
+#ifdef ZMQ_CPP11
 
 TEST_CASE("multipart codec empty", "[codec_multipart]")
 {
@@ -49,7 +49,7 @@ TEST_CASE("multipart codec big", "[codec_multipart]")
     mmsg.pushmem(big.data(), big.size());
     message_t msg = mmsg.encode();
     CHECK(msg.size() == 5 + 495);
-    CHECK(msg.data<unsigned char>()[0] == 0xFF);
+    CHECK(msg.data<unsigned char>()[0] == std::numeric_limits<uint8_t>::max());
     CHECK(msg.data<unsigned char>()[5] == 'X');
 
     CHECK(mmsg.size() == 1);
@@ -95,14 +95,14 @@ TEST_CASE("multipart codec encode too big", "[codec_multipart]")
 {
     using namespace zmq;
 
-    const size_t too_big_size = 1L + 0xFFFFFFFF;
-    CHECK(too_big_size > 0xFFFFFFFF);
+    const size_t too_big_size = 1L + std::numeric_limits<uint32_t>::max();
+    CHECK(too_big_size > std::numeric_limits<uint32_t>::max());
     char* too_big_data = new char[too_big_size];
     multipart_t mmsg(too_big_data, too_big_size);
     delete [] too_big_data;
 
     CHECK(mmsg.size() == 1);
-    CHECK(mmsg[0].size() > 0xFFFFFFFF);
+    CHECK(mmsg[0].size() > std::numeric_limits<uint32_t>::max());
 
     CHECK_THROWS_AS(
         mmsg.encode(),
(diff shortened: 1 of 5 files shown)


gummif:
Have you considered implementing these as free functions similar to send_multipart? That would allow constructing the message without creating a multipart_t object.

Also, is the protocol documented somewhere?

brettviren:
Hi @gummif.  I did consider it but seeing all the other helper methods in `multipart_t` that add value to `m_parts` I figured having `encode()/decode()` as methods fit into the current convention.

I agree what you say about a free function would not require a `multipart_t` to be created if the user doesn't need one.  OTOH, a free function means users that **do** want a `multipart_t` need an extra step.

What do you think about adding free functions in addition to the methods?  These functions could hold the actual codec code and be called by the methods.  Furthermore the functions could be templated to accept an iterator range so not tied to a particular container.

I haven't found anywhere in ZeroMQ ecosystem that documents the codec.  It's very simple and I figured it out by reading the few lines of `zmsg_encode()/_decode()` in CZMQ.   I do happen to have some blurb in the README of the package that currently drives the need for this codec and I reproduce it below:

> The encoding is compatible with the CZMQ `zmsg_encode()` function. Specifically, the data of small messages less than 255 bytes are prefixed with a 1-byte size. Larger messages are prefixed by a single byte `0xFF` followed by a 4-byte size value.

BTW, I have a Python implementation for this as well:

https://github.com/brettviren/generaldomo/blob/master/generaldomo/zhelpers.py#L75

gummif:
Free functions + member functions for convenience sounds good.

I wonder if this should be officially documented (in libzmq maybe), especially since the new SERVER/CLIENT socket types can only handle single part messages.

brettviren:
> Free functions + member functions for convenience sounds good

So, how would this additional feature be handled?  Does this PR go through and adding the free functions would be another PR (by you maybe?) or does this PR gain a second commit (by me) or do I need to withdraw this PR and make yet another?

gummif:
I guess you can choose. Either continue and  implement it in this PR, with new commits or rebasing, or that is a feature that can be added later in a second PR.

zmq_addon.hpp
```diff
@@ -469,6 +469,79 @@ class multipart_t
...
+    bool decode(const message_t& encoded) {
+        const unsigned char *source = encoded.data<unsigned char>();
+        const unsigned char *limit = source + encoded.size();
+        multipart_t decoded;
+        while (source < limit) {
+            size_t part_size = *source++;
+            if (part_size == 255) {
+                if (source > limit - 4) {
+                    // garbage, bail
+                    return false;
```
gummif: I wonder if throwing an exception (`std::argument_error` maybe) would be better, since that should be rare and not ignored.

tests/multipart.cpp
```diff
@@ -178,4 +178,39 @@ TEST_CASE("multipart legacy test", "[multipart]")
...
 }
+
+TEST_CASE("multipart codec", "[multipart]")
+{
+    using namespace zmq;
+
+    multipart_t mmsg;
+    mmsg.addstr("Hello World");
+    message_t msg = mmsg.encode();
+    assert(msg.size() == 1 + 11); // small size packing
```
gummif: The CHECK or REQUIRE macro should be used for unit tests.

Looks like the older unit tests in the file need a bit a refactoring.

sigiesec:
> Free functions + member functions for convenience sounds good.

Agree :)

> I wonder if this should be officially documented (in libzmq maybe), especially since the new SERVER/CLIENT socket types can only handle single part messages.

I think the appropriate place to specify this independently from the czmq implementation would be a ZeroMQ RFC (https://rfc.zeromq.org/). libzmq only provides the low-level API, which does not deal with this (which is why I was concerned about adding this to cppzmq, which wraps the low-level libzmq API). @bluca WDYT? Does it make sense to add a RFC for the multipart encoding?

zmq_addon.hpp
```diff
@@ -469,6 +469,79 @@ class multipart_t
...
+        for (auto& part : *this) {
+            size_t part_size = part.size();
+            size_t count_size = 5;
+            if (part_size < 255) {
+                count_size = 1;
+            }
+            mmsg_size += part_size + count_size;
+        }
+
+        message_t encoded(mmsg_size);
```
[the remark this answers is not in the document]
sigiesec: Hm, I don't think so. It's a corner case, but why should it be invalid to have a 0-part multi-part message?

But I don't know if `zmsg_encode` allows that. It should be done like that function does, since the goal is to be compatible.

@brettviren Please double-check and confirm this. In any case, please add a test case for this corner case.

zmq_addon.hpp
```diff
@@ -469,6 +469,79 @@ class multipart_t
...
+        unsigned char* buf = encoded.data<unsigned char>();
+        for (auto& part : *this) {
+            size_t part_size = part.size();
```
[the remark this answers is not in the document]
sigiesec: Yes, that should be checked. I think a run-time exception would be more appropriate here.

zmq_addon.hpp
sigiesec: I also think that an exception is preferrable for this case, in line with the rest of the API. If there are really use cases, where this might be a more common case, another `try_decode` function might be added that returns a bool in that case. In that case, the return value should be declared nodiscard, when available.

zmq_addon.hpp
```diff
@@ -469,6 +469,79 @@ class multipart_t
...
+            *buf++ = (part_size >> 24) & 255;
+            *buf++ = (part_size >> 16) & 255;
+            *buf++ = (part_size >>  8) & 255;
+            *buf++ =  part_size        & 255;
+            memcpy (buf, part_data, part_size);
+            buf += part_size;
+        }
+        return encoded;
+    }
+    // Decode the CZMQ zmsg_t/zframe_t compatible encoded single part
```
sigiesec: nit: please add a blank line before this.

bluca:
Perhaps - CZMQ until now has maintained a few protocols on its own, intended to be used only internally and thus not published. If someone wants to codify those, that should be fine - they change rarely, if ever.

zmq_addon.hpp
```diff
@@ -469,6 +469,79 @@ class multipart_t
...
+    // message into multi-parts and append to self.  Return true if
```
sigiesec: The appending behaviour is a bit disputable, and makes me question if having this as a non-static member function is a good idea. Isn't this more error-prone than useful? What about making this a static member function that returns a `multipart_t`?

If this stays like this, there should be at least one test that checks if it appends as expected.

zmq_addon.hpp
```diff
@@ -469,6 +469,79 @@ class multipart_t
...
+    // self was updated.  A false value will be returned if the
+    // encoded is garbage.
+    bool decode(const message_t& encoded) {
+        const unsigned char *source = encoded.data<unsigned char>();
+        const unsigned char *limit = source + encoded.size();
```
sigiesec: nit: could be declared top-level const
```suggestion
        const unsigned char * const limit = source + encoded.size();
```

tests/multipart.cpp
```diff
@@ -178,4 +178,39 @@ TEST_CASE("multipart legacy test", "[multipart]")
     assert(received.empty());
     assert(str == "One-hundred");
 }
+
+TEST_CASE("multipart codec", "[multipart]")
```
sigiesec: This seems to contain two test cases that somehow depend on each other. Please split this up, and consider this when adding further test cases.

zmq_addon.hpp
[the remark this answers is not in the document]
brettviren: So far I have yet to require sending an empty multipart-encoded in any higher-level protocols I've used.  Certainly, if it can be done it will be done eventually.

I think the proper thing is for an empty `multipart_t` to encode to the default (empty) `message_t` and vice versa. 

I'll double check what CZMQ does here and add some tests for this case.

zmq_addon.hpp
brettviren: Sounds good.  Consider this and the above added to my list.

tests/multipart.cpp
brettviren: I'll leave the older test to others but I will re-instrument this test with macros as suggested.

zmq_addon.hpp
brettviren: I don't fully understand this comment.  Are you worried the user may allow a `decode()` to append to a non-empty `multipart_t`?  If so I vacillated on the choice to allow it.  But, the opposite idiom is used in `multipart_t::recv()` so maybe I chose badly.  

If the comment is about something else, please elaborate.

tests/multipart.cpp
brettviren: Will do.  They are actually independent.  I'll add the others tests suggested as individual tests.

brettviren:
Thanks for all the comments.  One clarification requested and the rest all seem good to me.

zmq_addon.hpp
gummif: Yes, should it append or replace the content. Replacing is probably more common or expected behavior. Or as suggested use  `multipart_t mm = multipart_t::decode(my_message);`.

brettviren:
@gummif I've been able to make `encode()` a free function templated on a Range based on your `send_multipart()`.  To do similar for `decode()` I need a way to fill a `multipart_t` in a `std::` container looking manner.  I see I can mimic what you did with `recv_multipart()` it seems this function can not be used to directly fill a `multipart_t`. 

To make `recv_multipart()` and a free `decode()` work, I think `multipart_t` needs to be changed to look more like an STL container.  I'm almost at the edge of my C++ abilities, but I'll try to minimally add to `multipart_t` what's needed.  If you or others have any immediate suggestions, please let me know. 

I will appreciate another round of careful reviewing after I update the PR.

brettviren:
Turns out only minor additions to `multipart_t` are needed to let it be used with a `std::back_inserter` to provide `std::copy()` like calling for `decode()`.  I didn't try but I think this gives the knock-on benefit of letting `recv_multipart()` to actually be used with a `multipart_` (unless I miss this already working).

I removed the Boolean returns in favor of throwing.  I picked `std::range_error` if absurdly large parts are used (more than 32 bits) and `std::out_of_range` if encoded data is garbage and would otherwise lead to overflow errors.

I double checked what CZMQ does with empty multiparts and indeed it does what we do here.

I moved the codec tests to their own file and added more to cover the free functions empties, tickling the out of range exception.  I also made and ran a test that exercises the range error with a 4GB+ part but I disabled it as I worry about false negative failures when run on low-memory hosts (travis-ci?).

I also fixed the nits and ran clang-format.

And as I write this I see travis-ci is failing.... Of course it all is clean on my system.  Looks like a compilation error on one version of GCC and not related directly to the changes.  I'll fix this and see if I can squash that fix to what I just pushed to keep each commit in this PR green.

brettviren:
Yay, green.  Please have a look at 3191ae6 @sigiesec and @gummif and anyone else interested.

tests/multipart.cpp
```diff
@@ -178,4 +178,5 @@ TEST_CASE("multipart legacy test", "[multipart]")
     assert(received.empty());
     assert(str == "One-hundred");
 }
+
```
sigiesec: Stray change here, please remove this such that tests/multipart.cpp is not touched at all.

tests/codec_multipart.cpp
```diff
@@ -0,0 +1,168 @@
+#include <catch.hpp>
+#include <zmq_addon.hpp>
+
+#ifdef ZMQ_HAS_RVALUE_REFS
```
sigiesec: Do all of these tests require rvalue references?

zmq_addon.hpp
```diff
@@ -154,7 +155,122 @@ send_multipart(socket_ref s, Range &&msgs, send_flags flags = send_flags::none)
...
+template<class Range
+#ifndef ZMQ_CPP11_PARTIAL
+         ,
+         typename = typename std::enable_if<
+           detail::is_range<Range>::value
+           && (std::is_same<detail::range_value_t<Range>, message_t>::value
+               || detail::is_buffer<detail::range_value_t<Range>>::value)>::type
 #endif
+         >
+message_t encode(const Range &parts)
```
gummif: Do you think `encode_multipart` or `multipart_encode` (and `decode_multipart` or `multipart_decode`) would be more descriptive?

zmq_addon.hpp
```diff
@@ -154,7 +155,122 @@ send_multipart(socket_ref s, Range &&msgs, send_flags flags = send_flags::none)
...
+{
+    size_t mmsg_size = 0;
+
+    // First pass check sizes
+    for (const message_t &part : parts) {
```
gummif: To allow `const_buffer` and `mutable_buffer` to be used also, you can change this to `const auto&`, and use `static_cast<const unsigned char*>(part.data())` instead of `part.data<const unsigned char>()` below.

zmq_addon.hpp
```diff
@@ -154,7 +155,122 @@ send_multipart(socket_ref s, Range &&msgs, send_flags flags = send_flags::none)
...
+        size_t part_size = part.size();
+        if (part_size > 0xFFFFFFFF) {
```
gummif: Consider using `std::numeric_limits<uint32_t>::max()`

zmq_addon.hpp
```diff
@@ -154,7 +155,122 @@ send_multipart(socket_ref s, Range &&msgs, send_flags flags = send_flags::none)
...
+            // Size value must fit into uint32_t.
+            throw std::range_error("Invalid size, message part too large");
+        }
+        size_t count_size = 5;
+        if (part_size < 255) {
```
gummif: Consider using `std::numeric_limits<uint8_t>::max()`

zmq_addon.hpp
```diff
@@ -154,7 +155,122 @@ send_multipart(socket_ref s, Range &&msgs, send_flags flags = send_flags::none)
...
+
+    while (source < limit) {
+        size_t part_size = *source++;
+        if (part_size == 255) {
+            if (source > limit - 4) {
+                throw std::out_of_range(
+                  "Malformed encoding, overflow in reading size");
+            }
+            part_size =
+              (source[0] << 24) + (source[1] << 16) + (source[2] << 8) + source[3];
```
gummif: The types here are `unsigned char` and `int`, so I think there could be overflow if `source[0]` is big enough. There should probably by casts to `uint32_t`.

zmq_addon.hpp
```diff
@@ -154,7 +155,122 @@ send_multipart(socket_ref s, Range &&msgs, send_flags flags = send_flags::none)
...
+            source += 4;
+        }
+
+        if (source > limit - part_size) {
+            throw std::out_of_range("Malformed encoding, overflow in reading part");
+        }
+        message_t msg(source, part_size);
+        *out = std::move(msg);
```
gummif: `*out = message_t(source, part_size);` might be more efficient

zmq_addon.hpp
```diff
@@ -469,6 +590,23 @@ class multipart_t
         return true;
     }
 
+    // Return single part message_t encoded from this multipart_t.
+    message_t encode() const { return zmq::encode(*this); }
```
gummif: These new functions should be guarded by ZMQ_CPP11

tests/codec_multipart.cpp
gummif: This should probably be ZMQ_CPP11.

tests/multipart.cpp
brettviren: Oops!  Will do.

tests/codec_multipart.cpp
brettviren: Thanks, I'll check this.

zmq_addon.hpp
brettviren: I do not have a strong opinion. but I lean to leaving this as-is or maybe with a different `_qualifier`.  

I'm thinking about future types which may overload `encode()` where the label `_multipart` is not meaningful.  Eg, an application defined  `encode(const MyMessage& mm)` would not necessarily have any "multipartish".

If we had a good name for this codec then `encode_<codecname>()` (etc) would maybe be a better way to distinguish this.

zmq_addon.hpp
brettviren: Excellent, thanks.  I'll try this.

zmq_addon.hpp
brettviren: Indeed!  I confirm if `source[0]` is, eg `0xff` this will not give the expected value.

brettviren:
Latest commit should address all the comments from the previous one except one:  @gummif I still leave the free functions as `encode()/decode()` with my same rational I gave above.  If you feel strongly about using the longer names, please say so and I'll amend.  

Any other comments, @sigiesec or others?  I know it's been tedious to make (and address) them, but the iteration has made this small contribution into much higher quality than it started out as.

brettviren:
appveyor CI failed.  I know nothing about Windows development but it seems to not like the newly added `std::numeric_limits<>::max()`.  I realized I had not included `<limits>` explicitly.  It is maybe included implicitly on gcc but not on Windows?  But, I just force-pushed this fix and it did nothing to help.

I don't have any reasonable way to develop under Windows so I'm not sure what I should do.  Baring a better suggestion, I'll revert back to using literal `0xFF`, etc.

sigiesec:
> appveyor CI failed. I know nothing about Windows development but it seems to not like the newly added `std::numeric_limits<>::max()`. I realized I had not included `<limits>` explicitly. It is maybe included implicitly on gcc but not on Windows? But, I just force-pushed this fix and it did nothing to help.

The windows headers define macros named `min` and `max`... which is horrible. See http://www.suodenjoki.dk/us/archive/2010/min-max.htm, e.g. I am not completely sure, if we should define `NOMINMAX` in the CI script, or in `zmq.hpp` (when on Windows). Since it can't work without defining it (and almost all Windows projects sooner or later define it anyway), the latter seems to be a better idea. So 
```
#ifdef _WIN32
#ifndef NOMINMAX
#define NOMINMAX
#endif
#endif
```

I am not sure what exactly pulls `windows.h` in, so defining it at the very top because the first include directive is probably the best idea.

> 
> I don't have any reasonable way to develop under Windows so I'm not sure what I should do. Baring a better suggestion, I'll revert back to using literal `0xFF`, etc.

gummif:
I think the cause is the winsock2.h inlude in zmq.h (which has been removed on master is seems).

gummif:
Otherwise, looks good to me.

zmq_addon.hpp
brettviren: Support and tests for `const_buffer` and `mutable_buffer` is in the latest commit.

zmq_addon.hpp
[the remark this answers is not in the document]
brettviren: Indeed CZMQ encodes an empty multipart into an empty single part.  Tests for this case are added.

zmq_addon.hpp
brettviren: Looks like @sigiesec's suggestion for the windows problem let appveyor test pass.  So these use of limits in general looks good now.

brettviren:
Okay, looks all good after telling windows not to be silly about `max`.  I don't know how to clear the "changes requested" but I think everything was addressed.  Let me know if otherwise.

brettviren:
@sigiesec @gummif are you okay to merge this PR?

PR #392, opened 2020-03-07
Problem: no actor pattern.

brettviren:
Solution: provide one which is conceptually similar to CZMQ `zactor_t` in a new and optional header.

This is part of what was initially #388.

### DIFF
--- a/zmq_actor.hpp
+++ b/zmq_actor.hpp
@@ -27,93 +27,72 @@
 #include "zmq.hpp"
 
 #ifdef ZMQ_CPP11
-// Actor needs some threading.
-// Initial implementation uses std::thread so C++11 only.
+
 #include <thread>
 
-namespace zmq {
+namespace zmq
+{
+/*! Return two paired PAIR sockets.
+   
+  First is bound and second is connected via inproc://.
+*/
+std::pair<zmq::socket_t, zmq::socket_t> create_linked_pairs(context_t &ctx)
+{
+    std::pair<zmq::socket_t, zmq::socket_t> ret{socket_t(ctx, socket_type::pair),
+                                                socket_t(ctx, socket_type::pair)};
 
-    typedef std::pair<zmq::socket_t, zmq::socket_t> pipe;
+    std::stringstream ss;
+    ss << "inproc://paired-" << std::hex << ret.first.handle() << "-"
+       << ret.second.handle();
+    std::string addr = ss.str();
+    ret.first.bind(addr.c_str());
+    ret.second.connect(addr.c_str());
+    return ret;
+}
 
-    /*! Return a pair of PAIR sockets.
-     *
-     * First is bound and second is connected via inproc://
-     */
-    pipe create_pipe(context_t& ctx)
-    {
-        pipe ret{socket_t(ctx, socket_type::pair),
-                 socket_t(ctx, socket_type::pair)};
-
-        std::stringstream ss;
-        ss << "inproc://pipe-"
-           << std::hex
-           << ret.first.handle()
-           << "-"
-           << ret.second.handle();
-        std::string addr = ss.str();
-        ret.first.bind(addr.c_str());
-        ret.second.connect(addr.c_str());
-        return ret;
-    }
-
-
-    // This shim calls the actor function and then provides hard-wired
-    // protocol for shutdown, mimicking how CZMQ zactor_t does it.
+/*! Spawn a function in a thread and communicate over a link.
+     
+  The actor function must take a socket and zero or more optional
+  arguments such as:
+     
+      void func(zmq::socket sock, ...);
+      zmq::actor_t actor(ctx, func, ...);
+     
+  The socket passed in is one end of a bidirection link shared
+  with the application thread.  The application thread may get
+  the other end of the link by calling zmq::actor_t::link().
+     
+  Note: unlike CZMQ zactor_t, zmq::actor_t does not institute any
+  startup/shutdown protocol across the link.  If any is needed it
+  is left to the calling application.
+*/
+class actor_t
+{
+  public:
+    // Template constructor and not class to perform type erasure
+    // of the function type.
     template<typename Func, typename... Args>
-    void actor_shim(socket_t&& apipe, Func fn, Args... args) {
-        fn(apipe, args...);
-        apipe.send(zmq::message_t{}, zmq::send_flags::none);
+    actor_t(context_t &ctx, Func fn, Args... args)
+    {
+        socket_t asock;
+        std::tie(asock, _sock) = create_linked_pairs(ctx);
+
+        _thread = std::thread(
+          [fn = std::forward<Func>(fn)](zmq::socket_t asock, Args... args) {
+              fn(std::move(asock), std::forward<Args>(args)...);
+          },
+          std::move(asock), std::forward<Args>(args)...);
     }
 
-    /*! An CZMQ zactor_t like class.
-     *
-     * Construct it with an actor function and any args.  
-     *
-     * The function will be spawned in a thread.
-     *
-     * Application may use pipe() for a socket with which to
-     * communicate with the actor.
-     *
-     * The destructor will signal through the pipe and wait for the
-     * actor function to exit, if indeed it is still around.
-     */
-    class actor_t {
-    public:
+    // Waits for actor function to exit.
+    ~actor_t() { _thread.join(); }
 
-        template<typename Func, typename... Args>
-        actor_t(context_t& ctx, Func fn, Args... args) {
-            socket_t apipe;
-            std::tie(apipe, _pipe) = create_pipe(ctx);
-                              
-            _thread = std::thread(actor_shim<Func, Args...>,
-                                  std::move(apipe),
-                                  fn,
-                                  args...);
+    socket_ref link() { return _sock; }
 
-            zmq::message_t rmsg; // wait for actor ready signal
-            auto res = _pipe.recv(rmsg);
-        }
-
-        ~actor_t() {
-            auto sres = _pipe.send(message_t("$TERM",5), send_flags::dontwait);
-            if (sres) {
-                message_t rmsg;
-                auto res = _pipe.recv(rmsg, recv_flags::none);
-            }
-            _thread.join();
-        }
-
-        socket_t& pipe() { return _pipe; }
-        
-    private:
-        actor_t(const actor_t &) ZMQ_DELETED_FUNCTION;
-        void operator=(const actor_t &) ZMQ_DELETED_FUNCTION;
-
-    private:
-        socket_t _pipe;
-        std::thread _thread;
-    };
-    
+  private:
+    socket_t _sock;
+    std::thread _thread;
+};
 }
 
 #endif
--- a/zmq_actor.hpp
+++ b/zmq_actor.hpp
@@ -1,0 +1,120 @@
+/*
+    Copyright (c) 2020 ZeroMQ community
+    Copyright (c) 2020 Brett Viren
+
+    Permission is hereby granted, free of charge, to any person obtaining a copy
+    of this software and associated documentation files (the "Software"), to
+    deal in the Software without restriction, including without limitation the
+    rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+    sell copies of the Software, and to permit persons to whom the Software is
+    furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included in
+    all copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+    AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+    IN THE SOFTWARE.
+*/
+
+#ifndef __ZMQ_ACTOR_HPP_INCLUDED__
+#define __ZMQ_ACTOR_HPP_INCLUDED__
+
+#include "zmq.hpp"
+
+#ifdef ZMQ_CPP11
+// Actor needs some threading.
+// Initial implementation uses std::thread so C++11 only.
+#include <thread>
+
+namespace zmq {
+
+    typedef std::pair<zmq::socket_t, zmq::socket_t> pipe;
+
+    /*! Return a pair of PAIR sockets.
+     *
+     * First is bound and second is connected via inproc://
+     */
+    pipe create_pipe(context_t& ctx)
+    {
+        pipe ret{socket_t(ctx, socket_type::pair),
+                 socket_t(ctx, socket_type::pair)};
+
+        std::stringstream ss;
+        ss << "inproc://pipe-"
+           << std::hex
+           << ret.first.handle()
+           << "-"
+           << ret.second.handle();
+        std::string addr = ss.str();
+        ret.first.bind(addr.c_str());
+        ret.second.connect(addr.c_str());
+        return ret;
+    }
+
+
+    // This shim calls the actor function and then provides hard-wired
+    // protocol for shutdown, mimicking how CZMQ zactor_t does it.
+    template<typename Func, typename... Args>
+    void actor_shim(socket_t&& apipe, Func fn, Args... args) {
+        fn(apipe, args...);
+        apipe.send(zmq::message_t{}, zmq::send_flags::none);
+    }
+
+    /*! An CZMQ zactor_t like class.
+     *
+     * Construct it with an actor function and any args.  
+     *
+     * The function will be spawned in a thread.
+     *
+     * Application may use pipe() for a socket with which to
+     * communicate with the actor.
+     *
+     * The destructor will signal through the pipe and wait for the
+     * actor function to exit, if indeed it is still around.
+     */
+    class actor_t {
+    public:
+
+        template<typename Func, typename... Args>
+        actor_t(context_t& ctx, Func fn, Args... args) {
+            socket_t apipe;
+            std::tie(apipe, _pipe) = create_pipe(ctx);
+                              
+            _thread = std::thread(actor_shim<Func, Args...>,
+                                  std::move(apipe),
+                                  fn,
+                                  args...);
+
+            zmq::message_t rmsg; // wait for actor ready signal
+            auto res = _pipe.recv(rmsg);
+        }
+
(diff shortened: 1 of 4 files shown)


gummif:
I am not familiar wih this pattern. Can you explain what it is used for and provide 
example code for how you would use it?

brettviren:
Hi again @gummif.  The actor pattern is really useful.  There are some very trivial examples in the test but they don't give much flavor.  There's a lot that can be said about the actor pattern and I don't know how much it makes sense to relate here so here's a brief info dump of some of my experience with it:  

For me, the actor pattern is what turned ZeroMQ from merely being awesome to being a revolution in how I think about multi-threaded code.

It actually comes from a paper published way back in 1973!  But, I learned of it through ZeroMQ, specifically the `zactor_t` CLASS of CZMQ.  It's a basis for Zyre and all the clients and servers generated by Zproto (including my ZPerfMQ).  Hintjens has learning it as a lesson from one of his ten laws in this post:  http://hintjens.com/blog%3A100  which I definitely took to heart.

The repo in the other PR I mentioned is already using this PR's actor implementation, eg here is a test that uses it "in anger":

https://github.com/brettviren/generaldomo/blob/master/test/test_gdp.cpp#L34
https://github.com/brettviren/generaldomo/blob/master/test/test_gdp.cpp#L95

That first link goes to an actor function and the second starts some lines that create several actors.

The CZMQ implementation of the actor pattern plays a central role in a data acquisition subsystem I wrote:

https://github.com/brettviren/ptmp

It heavily exploits the actor pattern to allow construction of applications taking full advantage of the transport transparency of ZeroMQ sockets.  A single executable can consume configuration to make many actors in one program talking over `inproc://` or those same actors can be spread over many instances of this one general program and execute on many computers talking `tcp://`.  It allows one to scale out the overall application as needed without needing any new coding.

I'm in the process of developing a next generation version of that package and I want to base it on cppzmq instead of CZMQ.  Since the pattern didn't exist I gave a shot at implementing it and figured that if it's so useful to me it would probably be useful to some others, thus this PR.  

I hope this info is useful. :-)

gummif:
I can see the use for this class in C or before C++11. However with modern standard library features and libraries I do not see the need since there there are more efficient ways to do the same thing.

As I understand it, the actor is a class for a worker thread (with state) that talks to another thread via inproc sockets. So to talk to actors on other processes/machines you need another socket inside the actor thread?

I wonder if e.g. `boost::asio::thread_pool` (or some other `boost::asio::executor`)  will be sufficient if you need multiple actors without state (if per thread state is needed then thread ids could be mapped to state). If you need reply data then `std::future` or callbacks could be used. This of course requires careful thought about thread safety. 

I might be underestimating how powerful actors are or how widespread the usage of `zmq::message_t` is in your code.

zmq_actor.hpp
```diff
@@ -0,0 +1,120 @@
...
+     */
+    class actor_t {
+    public:
+
+        template<typename Func, typename... Args>
+        actor_t(context_t& ctx, Func fn, Args... args) {
+            socket_t apipe;
+            std::tie(apipe, _pipe) = create_pipe(ctx);
+                              
+            _thread = std::thread(actor_shim<Func, Args...>,
```
gummif: Why not just
```c++
 template<typename Func>
 actor_t(context_t& ctx, Func fn) {
    socket_t apipe;
     std::tie(apipe, _pipe) = create_pipe(ctx);
    _thread = std::thread([apipe = std::move(apipe)] {
        fn(apipe);
        apipe.send(zmq::message_t{}, zmq::send_flags::none);
    };
    ...
```

zmq_actor.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+            zmq::message_t rmsg; // wait for actor ready signal
+            auto res = _pipe.recv(rmsg);
+        }
+
+        ~actor_t() {
+            auto sres = _pipe.send(message_t("$TERM",5), send_flags::dontwait);
+            if (sres) {
+                message_t rmsg;
+                auto res = _pipe.recv(rmsg, recv_flags::none);
```
gummif: Is this required, is joining the thread not sufficient?

zmq_actor.hpp
```diff
@@ -0,0 +1,120 @@
...
+            }
+            _thread.join();
+        }
+
+        socket_t& pipe() { return _pipe; }
+        
+    private:
+        actor_t(const actor_t &) ZMQ_DELETED_FUNCTION;
+        void operator=(const actor_t &) ZMQ_DELETED_FUNCTION;
```
gummif: This is not requires since socket and thread are not copyable. Is the intent that the class is also not moveable? (because deleting the copy functions disables the generation of the move functions)

zmq_actor.hpp
```diff
@@ -0,0 +1,120 @@
...
+            socket_t apipe;
+            std::tie(apipe, _pipe) = create_pipe(ctx);
+                              
+            _thread = std::thread(actor_shim<Func, Args...>,
+                                  std::move(apipe),
+                                  fn,
+                                  args...);
+
+            zmq::message_t rmsg; // wait for actor ready signal
+            auto res = _pipe.recv(rmsg);
```
gummif: Is this necessary? If the actor fails and does not send a ready signal, and only sends the shutdown signal, will that lead to deadlocks?

zmq_actor.hpp
```diff
@@ -0,0 +1,120 @@
...
+#include "zmq.hpp"
+
+#ifdef ZMQ_CPP11
+// Actor needs some threading.
+// Initial implementation uses std::thread so C++11 only.
+#include <thread>
+
+namespace zmq {
+
+    typedef std::pair<zmq::socket_t, zmq::socket_t> pipe;
```
gummif: The nomenclature is a bit confusing, here pipe is a pair of two sockets, but in actor_t pipe() is a single socket.

zmq_actor.hpp
```diff
@@ -0,0 +1,120 @@
...
+        ret.first.bind(addr.c_str());
+        ret.second.connect(addr.c_str());
+        return ret;
+    }
+
+
+    // This shim calls the actor function and then provides hard-wired
+    // protocol for shutdown, mimicking how CZMQ zactor_t does it.
+    template<typename Func, typename... Args>
+    void actor_shim(socket_t&& apipe, Func fn, Args... args) {
```
gummif: This functions is not required, since we can use lambdas in C++11 (see comment below).

zmq_actor.hpp
brettviren: If it works, that's fine with me.  I used the external shim as I was being influenced by how CZMQ does it.

zmq_actor.hpp
brettviren: This `recv()` implements the minimal actor shutdown protocol that CZMQ implements.

zmq_actor.hpp
brettviren: Thanks, this was to stop copying.  The loss of move was not intended so these should be removed.

zmq_actor.hpp
brettviren: This is part of the minimal actor start up protocol that CZMQ implements.  The parent thread is supposed to wait until given the "go ahead" by the actor function.  In most cases, an immediate "go ahead" is all I've needed.  But, I figure CZMQ has it for a reason so I mimicked it.

zmq_actor.hpp
brettviren: I agree, it is imprecise.  I've seen "pipe" refer to the pair and to either end.  More often it's the latter, but that's also imprecise as "pipe" doesn't communicate well which end is being referred to.  I am not thinking of any better terms so I keep consistency with CZMQ.  Maybe `pipe_ends`.  Suggestions?

zmq_actor.hpp
brettviren: Yep, check.  

If you have any comments on pass by value vs pass by rvalue-reference, please give them.  I tried to make the template compile with `Func&& fn` and `Args&&... args` but could not.  This idiom is new to me so maybe it's wrong to try to use it here.

brettviren:
@gummif thank you very much for the comments, I added some responses.

zmq_actor.hpp
gummif: Pipes imply one directional flow of data (http://api.zeromq.org/2-1:zmq-socket#toc11), so I think that is not an optimal name (same as channel).

So maybe you could not provide this typedef at all, and instead call the creation function `create_socket_pair` or `create_paired_sockets`. I'm not sure about the `pipe()` function, maybe just `get_socket()`.

zmq_actor.hpp
gummif: OK, I will take a look an see if there is a rational for that.

zmq_actor.hpp
```diff
@@ -0,0 +1,120 @@
...
+        fn(apipe, args...);
+        apipe.send(zmq::message_t{}, zmq::send_flags::none);
+    }
```
sigiesec: I think this should work (and not copy the arguments):
```suggestion
    template<typename Func, typename... Args>
    void actor_shim(socket_t apipe, const Func &fn, Args&&... args) {
        fn(apipe, std::forward<Args>(args)...);
        apipe.send(zmq::message_t{}, zmq::send_flags::none);
    }
```
But as discussed below, probably the function isn't required at all.

sigiesec:
> I can see the use for this class in C or before C++11. However with modern standard library features and libraries I do not see the need since there there are more efficient ways to do the same thing.
> 
> As I understand it, the actor is a class for a worker thread (with state) that talks to another thread via inproc sockets. So to talk to actors on other processes/machines you need another socket inside the actor thread?
> 
> I wonder if e.g. `boost::asio::thread_pool` (or some other `boost::asio::executor`) will be sufficient if you need multiple actors without state (if per thread state is needed then thread ids could be mapped to state). If you need reply data then `std::future` or callbacks could be used. This of course requires careful thought about thread safety.
> 
> I might be underestimating how powerful actors are or how widespread the usage of `zmq::message_t` is in your code.

I think this is in some way an alternative design to a thread pool as implemented by `boost::asio`. In general, it is probably not easy to trade one for the other. Since czmq and applications based on it make wide-spread use of it, I guess it's safe to assume that it has its use cases, without digging deeper into it (I have never used that model, admittedly).

@gummif has already provided an extensive review, I have nothing to add to that. @brettviren, I will merge it once @gummif has approved the changes. But again, please ensure to run clang-format eventually. Thanks!

brettviren:
Thanks @sigiesec.  I will implement the suggestions from you and @gummif and append a second commit to this PR for review.  It may be a few days as the work week is already inserting other tasks in my queue.

One comment about use of thread pool: in my understanding it is important for ZeroMQ actor pattern to **not** employ a thread pool for running the actor function (at least not one that migrates the function to different threads over its lifetime).  This is so that any "non-thread-safe" sockets may correctly work inside the actor function.  This understanding is in the context of my recent questions on zeromq-dev mailing list about use of DEALER inside TBB flow graph / thread pool.  Any additional wisdom on this aspect is very welcome.

gummif:
All sockets used by multiple threads need to be guarded by a mutex (or other equivalent synchronization). But it is also possible to have "thread local" sockets, that is, each thread in the pool could request a socket or some state that is  only used by that thread. This can be accomplished using a mutex and a `std::map<std::thread_id, ThreadState>` for example.

zmq_actor.hpp
gummif: I believe this should be omitted. If a user needs some "go ahead" functionality, then that can be implemented easily in the user code.

zmq_actor.hpp
gummif: I think this blocking recv should be omitted also. I also think the shutdown message signal should be a empty message, to be consistent with the thread shutdown signal, and also similar to closing a ZMQ_STREAM connection. That would simply be
```c++
try {
  _pipe.send(zmq::const_buffer{}, send_flags::dontwait);
} catch (...) {} // catch exception in case of context shutdown
_thread.join();
```

brettviren: Hi @gummif.  Changing this mini-protocol will mean CZMQ actor functions can not be used as cppzmq actor functions.  Maybe there is a way to make an optional adapter function to provide a compatibility layer.  If so, I see no objection to this and your other suggestion and they will make the base code simpler. I'll look into it.  Thanks.

zmq_actor.hpp
brettviren: I got rid of the shim in favor of a lambda.

zmq_actor.hpp
brettviren: I use "link" and "linked pairs" nomenclature.

zmq_actor.hpp
brettviren: I've removed these.

zmq_actor.hpp
brettviren: For this and the other comments regarding emulating CZMQ's actor startup/shutdown mini-protocol, I've come to accept the minimalist approach you suggest.  I think the application will need to invent something for practical purposes but that is maybe best left to the user.  The only downside is that the shutdown part is best put in an actor destructor. 

I did add an example of implementing this at the application level in the tests.

This then leaves `zmq::actor_t` as just providing a bundling of thread/func/link and providing the type erasure of the func.

zmq_actor.hpp
brettviren: Thanks, I've done this with the addition of still allowing variatic Args.

brettviren:
I think I've addressed all the comments.  But, I now have one of my own:

With the CZMQ `zactor_t` style startup/shutdown mini-protocol now removed (which I agree now is best for cppzmq layer), this implementation of the actor pattern becomes rather minimal.  I think the main value left is letting the user not have to worry about getting all the various move/forward and templating right (which I hope are now correct).  

In a real/practical application I think some shutdown protocol is needed.  So, I now wonder if this minimal implementation is really worth adding as a "first class" feature in cppzmq.  Maybe it should better be relegated to an example which application developers can then take and customize?

I'm okay with either decision.  So, please contemplate.  Thanks!

zmq_actor.hpp
```diff
@@ -0,0 +1,99 @@
...
+                                                socket_t(ctx, socket_type::pair)};
+
+    std::stringstream ss;
+    ss << "inproc://paired-" << std::hex << ret.first.handle() << "-"
+       << ret.second.handle();
+    std::string addr = ss.str();
+    ret.first.bind(addr.c_str());
+    ret.second.connect(addr.c_str());
+    return ret;
+}
```
gummif: May I suggest a more generic function?
```C++
void link_inproc(socket_ref first, socket_ref second)
{
    std::stringstream ss;
    ss << "inproc://paired-" << std::hex << first.handle() << "-"
       << second.handle();
    std::string addr = ss.str();
    first.bind(addr.c_str());
    second.connect(addr.c_str());
}
```

gummif: That gives the user the option of setting socket options before binding/connecting.

zmq_actor.hpp
```diff
@@ -0,0 +1,99 @@
...
+    {
+        socket_t asock;
+        std::tie(asock, _sock) = create_linked_pairs(ctx);
+
+        _thread = std::thread(
+          [fn = std::forward<Func>(fn)](zmq::socket_t asock, Args... args) {
+              fn(std::move(asock), std::forward<Args>(args)...);
+          },
+          std::move(asock), std::forward<Args>(args)...);
+    }
```
gummif: I feel like this constructor is doing too much and can be simplified
```C++
    template<class Func>
    actor_t(context_t &ctx, Func&& fn)
    {
        socket_t asock;
        std::tie(asock, _sock) = create_linked_pairs(ctx);

        _thread = std::thread(
          [sock = std::move(sock), fn = std::forward<Func>(fn)]() mutable {
              fn(std::move(asock));
              // NOTE a shutdown signal would make this more useful
          });
    }
```

zmq_actor.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+    // Waits for actor function to exit.
+    ~actor_t() { _thread.join(); }
```
gummif: I see you have remove the thread and actor shutdown signalling. Is there a reason for that? How can deadlocks be avoided when the actor is destucting?

zmq_actor.hpp
brettviren: This is fine.  This function is needed (for now) only inside the `actor_t`.  I guess it could just as well be removed with the body residing in the `actor_t`.  Separating it to another function is influence from CZMQ.

zmq_actor.hpp
brettviren: This does two unwanted things: 

1) It removes the ability to pass arguments into the function.  I guess this can be replaced by requiring the user to provide a lambda.

2) It makes the `actor_t` typed.  One use case I have is to maintain a collection of actors which may have similar protocol on the "link" but which have different actor functions.

zmq_actor.hpp
brettviren: Maybe I misunderstood, but I removed this based on one of your previous comments.

brettviren:
I'm starting to think there are many ways users would want to make an actor pattern to suit their particular needs and that no single one is "correct" enough to properly put in this low-level package.  Maybe some future CZMQ++ package would be more suitable.

Are there objections if I close this PR and not pursue a merge?  (I'm also getting a bit too busy - who said WFH is less work!)

zmq_actor.hpp
gummif: 1. Lambdas are the way to go, but you could use `std::bind` or `std::bind_front` to construct the function object if that simplifies the code.
2. I don't think so. You can still have a `std::vector<actor_t>` with different functions. They are type erased in `std::thread`.

zmq_actor.hpp
gummif: Ok sorry about that. My opinion was to remove the startup protocol and keep the shutdown, but maybe simplify the shutdown and make it symmetric.

zmq_actor.hpp
brettviren: About 2.  Yes, you are right!  Sorry, my lack of reading comprehension. I misunderstood the example thinking you were suggesting to template the class.  I will give both of these a try.

zmq_actor.hpp
brettviren: Okay, I'll try this.  

About simplifying the shutdown: I can see the wisdom in sending a shutdown message with an explicit content (eg the `$TERM` used by CZMQ).  OTOH, if the shutdown message is generic it removes the ability to use that generic message as part of some application protocol with the actor. 

I guess I could make the content of the shutdown message be something passed into the `actor_t` so that the application can arrange a coordination between it and the actor function.  Kind of a shared "secret" key.  It could then default to being an empty message.  

I'll try this and your other suggestions and add a commit.  It might be a few days before I get to that.

Thanks for your continued input!

gummif: That sounds good. Maybe it is better to have it public, so that the "consumer" of the actor can easily verify that a message it receives is a shutdown signal, as in `const message_t& shutdown_signal() const`.

sigiesec:
> I'm starting to think there are many ways users would want to make an actor pattern to suit their particular needs and that no single one is "correct" enough to properly put in this low-level package. Maybe some future CZMQ++ package would be more suitable.

That might be an option :)
 
> Are there objections if I close this PR and not pursue a merge? (I'm also getting a bit too busy - who said WFH is less work!)

Sure. It's not lost. The PR can be reopened any time, if you come to the conclusion cppzmq actually is the right place.

Sorry I am also not that active these days. Take care :)

sigiesec:
As discussed, closing this for now. Feel free to re-open when picking this up again!

PR #393, opened 2020-03-15
Problem: Socket options lack type-safety 

gummif:
Solution: Implement a type for each socket option.

Each option has an associated type, therefore they
can't be simply defined as an enum class.

That's a lot of options! I think this is one of the last places that are truly dangerous in the current API, where e.g. `s.setsockopt(ZMQ_ROUTING_ID, std::string("foo"));` will not do what you expect.

### DIFF
--- a/tests/socket.cpp
+++ b/tests/socket.cpp
@@ -69,9 +69,9 @@ TEST_CASE("socket options", "[socket]")
     socket.set(zmq::sockopt::routing_id, "foobar");
     socket.set(zmq::sockopt::routing_id, zmq::buffer(id));
     socket.set(zmq::sockopt::routing_id, id);
-    #ifdef ZMQ_CPP17
+#ifdef ZMQ_CPP17
     socket.set(zmq::sockopt::routing_id, std::string_view{id});
-    #endif
+#endif
 
     std::string id_ret(10, ' ');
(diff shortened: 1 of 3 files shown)


gummif:
Yes, this is a bit of a burden but worth it I think since the package advertises type safety.
There is one benefit to this design that the user could define new or old deprecated options simply as `using old_opt = zmq::sockopt::integral_option<ZMQ_OLD_OPT, int>;` and use that to set and get options.

I think I could add the tests, shouldn't be a lot of trouble.

I did not clang format, I will do that.

sigiesec:
> Yes, this is a bit of a burden but worth it I think since the package advertises type safety.

Yes, you're right. It matches the design goals, and while it is an additional maintenance burden, it's worth it :)

> There is one benefit to this design that the user could define new or old deprecated options simply as `using old_opt = zmq::sockopt::integral_option<ZMQ_OLD_OPT, int>;` and use that to set and get options.

Yes, that's great!

> I think I could add the tests, shouldn't be a lot of trouble.

Ok, fine then :)

> I did not clang format, I will do that.

:)

gummif:
I was unable to test the bindtodevice and gssapi options. I also found a bug in libzmq documentation (https://github.com/zeromq/libzmq/pull/3851).

sigiesec:
Great, thanks a lot for doing this! I think missing the bindtodevice and gssapi options is acceptable :)

Issue #396, opened 2020-03-17
message_t::to_string inserts a trailing null char and breaks string comparisons


dpellegr:
Text messages sent via zmq get a trailing null char on the receiver side. The method message_t::to_string[_view] return a string size which is one byte larger than expected.
Comparison between such returned strings are broken unless one manually constructs a longer string and then sets the last char to null.

```
// req.cpp

#include <string>
#include <iostream>
#include <zmq.hpp>

int main() {
  zmq::context_t ctx;
  zmq::socket_t sock(ctx, zmq::socket_type::req);
  sock.connect("ipc:///tmp/zmqsocket");
  zmq::message_t msg{"abc"};
  sock.send(msg);
  sock.recv(msg);
  std::cout << msg.to_string_view() << std::endl;
  return 0;
}

```

```
// rep.cpp

#include <string>
#include <iostream>
#include <zmq.hpp>

int main() {
  zmq::context_t ctx;
  zmq::socket_t sock(ctx, zmq::socket_type::rep);
  sock.bind("ipc:///tmp/zmqsocket");
  zmq::message_t msg;
  while (true) {
    std::cout << "listening..." << std::endl;
    sock.recv(msg);
    // msg.to_string_view() returns "abc\0", therefore msg.to_string_view() != "abc"
    // note that the trailing null char eludes printing functions

    // When constructing strings C++ optimizes away trailing null chars
    // One needs to build a longer string and set the last char to null later
    std::string s{"abcX"}; 
    s[3] = '\0';

    // now the comparison works
    if ( msg.to_string_view() == s) {
      sock.send(zmq::buffer("def"));
    } else {
      sock.send(zmq::buffer("???"));
    }
  }
  return 0;
}

```

PR #397, opened 2020-03-17
message_t::to_string now produces the correct string length

dpellegr:
This fixes the creation of malformed std::strings including or going beyond the cstring null char termination. See https://github.com/zeromq/cppzmq/issues/396

Here is a minimal example reproducing the problem:
```
#include <string>
#include <cstring>
#include <iostream>

using namespace std;

int main() {
  char cstr[128];
  strcpy(cstr, "Hello! How are you?");

  std::string str(cstr, 128);
  cout << str.size() << " --- " << str << endl;

  return 0;
}
```
The output can be:
`128 --- Hello! How are you?PS���T���T�...`

The solution looks first for the terminating null char and then proceeds to copy up to that point. If the null char is not found the whole buffer is copied, as before.

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -556,13 +556,17 @@ class message_t
     // interpret message content as a string
     std::string to_string() const
     {
-        return std::string(static_cast<const char *>(data()), size());
+        const auto start = data<const char>();
+        const auto length = std::distance(start, std::find(start, start+size(), '\0'));
+        return std::string(start, length);
     }
 #ifdef ZMQ_CPP17
     // interpret message content as a string
     std::string_view to_string_view() const noexcept
     {
-        return std::string_view(static_cast<const char *>(data()), size());
+        const auto start = data<const char>();
+        const auto length = std::distance(start, std::find(start, start+size(), '\0'));
+        return std::string_view(start, length);
     }
 #endif
 


gummif:
You are correct in the issue that the behavior is inconsistent. But I think the problem is that the constructor for message_t is too greedy and changing it would be a breaking change.
One solution is to use the str_buffer function.

dpellegr:
Yeah, I didn't go too deep into the root cause. But I lost a couple of hours understanding why the string returned by to_string failed to compare to a literal and I just wanted to save the same pain to others. It is fine to reject the pull request, but, at least, I think that the to_string function should be marked unsafe (ot requiring special attention) clearly explaining this issue.

gummif:
Yes, this should also be better documented that messages and buffers from raw literals (and char[]) will contain everything including the null.

I would prefer you add new functions `to_string_no_null` (or similar) that do what you implemented here. Then `to_string` does exactly what it advertises (convert to a string without manipulation), and `to_string_no_null` can be used when the message may contain a terminating null.

dpellegr:
I think that if you copy to `std::string` then you want a proper string without null chars and other rubbish within its `size()`. If you want to copy the whole buffer, than a `std::vector<char>`, while structurally very similar, is definitely a better choice. So you could have `to_string` as I proposed and `to_vector` as it was before, without introducing `very_long_and_annoying_function_names` :)

gummif:
That would introduce non-zero overhead for most users. I would then argue that message_t should never be constructed from string literals `zmq::message_t("foo")`, but instead always from string or similar types, like `zmq::message_t("foo"sv)`,  `zmq::message_t(std::string_view{"foo"})` or  `zmq::message_t(zmq::str_buffer("foo"))`.

Maybe the greedy range constructor should be marked as deprecated, and overloads for string-like types (`const char*`, `std::string`, `std::string_view`) and `zmq::const_buffer` should be added instead. That would make the API safer to use and avoids accidental misuse.

sigiesec:
Is this obsoleted by #399?

gummif:
Yes I believe so.

sigiesec:
Closing this for now, as #399 will probably fix the issues.

PR #399, opened 2020-04-03
Problem: message_t ctor string inconsistency 

gummif:
Solution: Constructor taking generic ranges including string literals includes the null terminated char in
the message. Deprecate function and add overloads for strings.

This fixes the root cause of a problem observed in https://github.com/zeromq/cppzmq/issues/396 and discussed here https://github.com/zeromq/cppzmq/pull/397

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -348,6 +348,19 @@ inline std::tuple<int, int, int> version()
     zmq_version(&std::get<0>(v), &std::get<1>(v), &std::get<2>(v));
     return v;
 }
+
+#if !defined(ZMQ_CPP11_PARTIAL)
+namespace detail{
+template<class T> struct is_char_type
+{
+    // true if character type for string literals in C++11
+    static constexpr bool value =
+      std::is_same<T, char>::value || std::is_same<T, wchar_t>::value ||
+      std::is_same<T, char16_t>::value || std::is_same<T, char32_t>::value;
+};
+}
+#endif
+
 #endif
 
 class message_t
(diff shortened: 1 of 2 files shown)


sigiesec:
Sorry I am lagging behind somewhat. I will try to give this and your other PRs a closer look in the next days. Is there one particularly urgent?

gummif:
Alright, nothing urgent, just minor fixes.

sigiesec:
I still have some problems digesting this rather large PR. Unfortunately, the commits are not self-contained, as some of them fix issues from earlier commits. Is it possible to break this up into some self-contained steps, and rewrite the commits accordingly, in a way that would simplify review?

Is all of this really necessary to fix the issue from #396? (Not saying that the changes shouldn't be done, I am just trying to understand if there is one part fixing #396 and more things that are additional improvements)

gummif:
I am going to split the unrelated changes into other PRs (see e.g. https://github.com/zeromq/cppzmq/pull/416) before the rewrite.

gummif:
This has been simplified a lot and is ready for another review.

zmq.hpp
```diff
@@ -399,16 +399,36 @@ class message_t
...
+    // when called with a string literal.
+    // An overload taking const char* can not be added because
+    // it would be preferred over this function and break compatiblity.
     template<class Range,
              typename = typename std::enable_if<
                detail::is_range<Range>::value
                && ZMQ_IS_TRIVIALLY_COPYABLE(detail::range_value_t<Range>)
                && !std::is_same<Range, message_t>::value>::type>
+    ZMQ_DEPRECATED(
+        "from 4.7.0, use constructors taking iterators, (pointer, size) or strings instead")
```
sigiesec, line 414: I think in the future with C++20 ranges, it will seem odd to recommend use of an overload accepting a pair of iterators rather than a range. Could we add an overload `template<size_t N, > explicit message_t(const char (&rng)[N])` and deprecate just that?

sigiesec (changes requested):
Thanks, now it was must easier to review this :)

zmq.hpp
gummif: Yes good point.

sigiesec: Thanks!

sigiesec (approved):

PR #401, opened 2020-04-05
Problem: UB in message_t constructor

gummif:
Solution: Add check to guard against passing null to memcpy

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -365,7 +365,12 @@ class message_t
         int rc = zmq_msg_init_size(&msg, size_);
         if (rc != 0)
             throw error_t();
-        memcpy(data(), data_, size_);
+        if (size_)
+        {
+            // this constructor allows (nullptr, 0),
+            // memcpy with a null pointer is UB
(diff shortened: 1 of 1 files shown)


zmq.hpp
```diff
@@ -365,7 +365,12 @@ class message_t
         int rc = zmq_msg_init_size(&msg, size_);
         if (rc != 0)
             throw error_t();
-        memcpy(data(), data_, size_);
+        if (size_)
+        {
+            // this constructor allows (nullptr, 0),
```
sigiesec, line 370: Does the constructor allow that? Should it?

We can add the check to be on the extremely safe side anyway. (Does this actually cause UB on any existing implementation? I know the standard allows ot to be UB.)

gummif: I think it is common for constructing an empty message (in generic code). There is a similar check in libzmq in `zmq_recv`.

It will optimize away any subsequent null checks on gcc, see https://godbolt.org/z/vyx6eG, but not on clang.

PR #404, opened 2020-04-18
Problem: No endian check in encoding

gummif:
Solution: Always write message part sizes
using network order.


gummif:
If I am not mistaken, this has not been released yet. Is it considered breaking even if the functionality is only on master?

sigiesec:
Ah, sorry. Thanks for pointing out that this was only a recent addition. Then, it's fine of course. Thanks for catching this!

PR #405, opened 2020-04-19
Problem: Active poller double add mutates handler

gummif:
Solution: Check if socket already added before storing.
Add requirement that the handler is non-null. Improve
checks for zmq error codes.

### DIFF
--- a/zmq_addon.hpp
+++ b/zmq_addon.hpp
@@ -675,20 +675,17 @@ class active_poller_t
     {
         if (!handler)
             throw std::invalid_argument("null handler in active_poller_t::add");
-        auto it = decltype(handlers)::iterator{};
-        auto inserted = bool{};
-        std::tie(it, inserted) = handlers.emplace(
+        auto ret = handlers.emplace(
           socket, std::make_shared<handler_type>(std::move(handler)));
+        if (!ret.second)
+            throw error_t(EINVAL); // already added
         try {
-            base_poller.add(socket, events,
-                            inserted && *(it->second) ? it->second.get() : nullptr);
-            need_rebuild |= inserted;
+            base_poller.add(socket, events, ret.first->second.get());
+            need_rebuild = true;
         }
-        catch (const zmq::error_t &) {
+        catch (...) {
             // rollback
-            if (inserted) {
-                handlers.erase(socket);
-            }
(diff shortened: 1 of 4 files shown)


gummif:
Fixed

zmq_addon.hpp
```diff
@@ -675,20 +675,24 @@ class active_poller_t
...
         if (!handler)
             throw std::invalid_argument("null handler in active_poller_t::add");
-        auto it = decltype(handlers)::iterator{};
-        auto inserted = bool{};
-        std::tie(it, inserted) = handlers.emplace(
-          socket, std::make_shared<handler_type>(std::move(handler)));
+        auto it = handlers.find(socket);
+        if (it != handlers.end()) {
+            // add to poller to make in throw the expected error code
+            base_poller.add(socket, events, nullptr);
```
sigiesec, line 681: I don't completely understand the comment and why `base_poller.add` is called. Why can't we throw the expected exception directly from here?

gummif: Good question. What I would like to do is set errno and then throw error_t. But Im not sure what is the best platform independent way to set errno is, similar to zmq_errno.

This is so so that double add throws the same kind of exception as in poller_t, but that may not be required if too hard to implement properly.

zmq_addon.hpp
```diff
@@ -675,20 +675,24 @@ class active_poller_t
...
+            assert(false);
+            // fallback to runtime_error
+            base_poller.remove(socket);
+            throw std::runtime_error("socket already added");
```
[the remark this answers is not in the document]
gummif: This would really only run in case of buggy code. But might as well terminate instead to make it clearer.

zmq_addon.hpp
sigiesec: If the sole purpose is to get the right errno into `error_t` why not add a constructor to `error_t` that accepts an error number? We don't need to set `errno` at all.

zmq_addon.hpp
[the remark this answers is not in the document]
sigiesec: ... in case of a bug in which code? In cppzmq or in the application code? In the former case, an `assert` seems to be more appropriate to me.

PR #408, opened 2020-05-04
Problem: implicit conversion operators in context

gummif:
Solution: add handle() and mark operators as deprecated

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -788,6 +788,9 @@ class context_t
 
     ZMQ_EXPLICIT operator void const *() const ZMQ_NOTHROW { return ptr; }
 
+    ZMQ_NODISCARD void *handle() ZMQ_NOTHROW { return ptr; }
+
+    ZMQ_DEPRECATED("from 4.7.0, use handle() != nullptr instead")
     operator bool() const ZMQ_NOTHROW { return ptr != ZMQ_NULLPTR; }
 
     void swap(context_t &other) ZMQ_NOTHROW { std::swap(ptr, other.ptr); }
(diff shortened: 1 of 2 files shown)


zmq.hpp
```diff
@@ -784,10 +784,17 @@ class context_t
...
     //  Normally you should never need to use this.
+    ZMQ_DEPRECATED("from 4.7.0, use handle() instead")
     ZMQ_EXPLICIT operator void *() ZMQ_NOTHROW { return ptr; }
 
+    ZMQ_DEPRECATED("from 4.7.0, use handle() instead")
     ZMQ_EXPLICIT operator void const *() const ZMQ_NOTHROW { return ptr; }
 
+    ZMQ_NODISCARD void *handle() ZMQ_NOTHROW { return ptr; }
+
+    ZMQ_NODISCARD const void *handle() const ZMQ_NOTHROW { return ptr; }
```
sigiesec, line 795: I don't think having a const & non-const overload makes sense. What should the `const`  overload be used for? `libzmq` doesn't have a noting of a `const void*` context.

Technically, the const overload could return a `void*`. It would require a subtle philosophical discussion of logical constness to figure out if that is right. I guess I am fine with either solution, i.e. either a `const` or a non-`const` `handle` function.

gummif: OK, I'm fine with un-deprecating the explicit operators.

But regarding the const and non-const overloads, I agree the only the non-const is needed. But if we deprecate the implicit bool operator, then what is the expected way to check that a context is valid/open if we have a const reference to a context? I guess in that (rare) case one could use `static_cast<const void*>(ctx)` instead of `ctx.handle()`.

sigiesec: I think for `context_t`, this is not really relevant. `context_t` has no other `const` member functions, so a `const context_t&` is pretty useless. Or do you have a specific use case in mind that I am not aware of?

gummif: No, no particular use case that cant be easily solved with other means.

PR #412, opened 2020-05-10
Added support to std::array to poll()

Critical2104:
Added support to std::array to poll() (in addition to std::vector) to enhance C++/STL compatibility.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,6 +82,19 @@ matrix:
       env:
         - MATRIX_EVAL="CC=gcc-7 && CXX=g++-7" ENABLE_DRAFTS=ON
 
+    # Clang 8.0, draft disabled (default), latest libzmq (default), C++17
+    - os: linux
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+            - llvm-toolchain-trusty-8
+          packages:
+            - g++-9
+            - clang-8
+      env:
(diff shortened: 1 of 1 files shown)


CommanderBubble:
would it be worth adding support for iterators instead / as well as this, instead of adding support for each container explicitly?

zmq.hpp
```diff
@@ -313,6 +313,19 @@ inline int poll(std::vector<zmq_pollitem_t> &items,
...
 {
     return poll(items.data(), items.size(), timeout_);
+}		 
+
+template<std::size_t SIZE>
+inline int poll(std::array<zmq_pollitem_t, SIZE>& items, long timeout_ = -1)
+{
+    return poll(items.data(), items.size(), timeout_);
+}
+
```
sigiesec, line 323: Please remove this. The corresponding overload for `std::vector` and a `long` timeout is already deprecated. We shouldn't add a new one.

Critical2104: Not a problem removing that overload but I don't see it deprecated for `std::vector` on master

sigiesec: Oh, you're right. I confused that with the `const` overload. Anyway, it *should* be deprecated, and so it would be great not to add a similar overload :)

Critical2104: I amended the PR removing the said overload. Additionally I marked the std::vector one, taking long, as deprecated

sigiesec:
Thanks for addressing the comments.

Sorry, I tried to fix the history of your branch, but apparently the PR got closed and I cannot update it anymore. I will open a new PR from my fork with the changes.

Issue #381, opened 2020-01-27
void monitor_t::abort() is violating RAII principle


lp35:
Hi,

I run into a strange bug today, due to monitor_t.

To my understanding of the class, what init() does is reversed by abort() call.

abort() should then free the socket, to allow the zmq context to be freed.

In this case, abort should contains the following code:

```
       if (socketPtr)
            zmq_socket_monitor(socketPtr, NULL, 0);

        socketPtr = NULL;

        if (monitor_socket)
            zmq_close(monitor_socket);
```

The socket cannot be closed until the destructor is called, that can be problematic for some use cases.

I guess that issue #185 is related to this bug.

lp35:
(I can PR if needed)

sigiesec:
I am not sure if I understand the problem here completely. Shutting down monitoring right is rather tricky, unfortunately. I am currently investigating an issue in libzmq related to that as well. Could you share a test case that demonstrates the problem?

Issue #109, opened 2017-01-18
Poll monitoring sockets


v1bri:
I would like to be able to use `zmq::poll()` for normal ZMQ sockets and monitoring ZMQ sockets, to handle them all within a single thread. However this isn't currently possible with the `zmq::monitor_t` class due to the blocking nature of the `zmq::monitor_t::monitor()` method and the lack of an accessor for the socket that is monitoring `socketPtr`.

Could we augment the `zmq::monitor_t` class with the functionality to allow polling? Perhaps something like the following would work (apologies in advance for syntax)...

```c++
class monitor_t {
    // ...
    inline operator void* () ZMQ_NOTHROW
    {
        return monitorPtr;
    }

    inline operator void const* () const ZMQ_NOTHROW
    {
        return monitorPtr;
    }

    void monitor_start(socket_t &socket, const char *addr_, int events = ZMQ_EVENT_ALL)
    {
        int rc = zmq_socket_monitor(socket.ptr, addr_, events);
        if (rc != 0)
            throw error_t ();

        socketPtr = socket.ptr;
        monitorPtr = zmq_socket (socket.ctxptr, ZMQ_PAIR);
        assert (monitorPtr);

        rc = zmq_connect (monitorPtr, addr_);
        assert (rc == 0);

        on_monitor_started();
    }

    void monitor_poll()
    {
        zmq_msg_t eventMsg;
        zmq_msg_init (&eventMsg);
        rc = zmq_msg_recv (&eventMsg, s, 0);
        if (rc == -1 && zmq_errno() == ETERM)
            break;
        assert (rc != -1);

        // ...
    }

    void monitor(socket_t &socket, const char *addr_, int events = ZMQ_EVENT_ALL)
    {
        monitor_start(socket, addr_, events);
        while (true) {
            monitor_poll();
        }
    }
    // ...

private:
    void* socketPtr;
    void* monitorPtr;
};
```
This would allow a usage similar to the following...
```c++
void Worker(zmq::socket_t* raw_zmq_data_sock)
{
    std::unique_ptr<zmq::socket_t> zmq_data_sock(raw_zmq_data_sock);
    std::unique_ptr<zmq::monitor_t> zmq_monitor_sock(new zmq::monitor_t);

    zmq_monitor_sock->start_monitor(*zmq_data_sock, "abc");

    std::vector<zmq::pollitem_t> poll_items = {
[...]

PR #414, opened 2020-05-14
Feat/monitor poll

lp35:
This MR improve monitor_t class by addressing #381 and #109.

This lead to better application design and allows the same thread to do both monitoring and retrieving messages from the monitored socket, thus avoiding clunky code with threads joining everywhere.

### DIFF
--- a/tests/monitor.cpp
+++ b/tests/monitor.cpp
@@ -156,6 +156,24 @@ TEST_CASE("monitor init get event count", "[monitor]")
         }
     }
 
+    SECTION("poller_t get_event")
+    {
+        zmq::poller_t<> poller;
+        CHECK_NOTHROW(poller.add(monitor, zmq::event_flags::pollin));
+
+        while (total < expected_event_count)
+        {
+            std::vector<zmq::poller_event<>> events(1);
+            if(0 == poller.wait_all(events, std::chrono::milliseconds{ 100 }))
+                continue;
+
+            CHECK(zmq::event_flags::pollin == events[0].events);
+            CHECK(monitor.get_event(eventMsg, address));
+
+            lbd_count_event(eventMsg);
+        }
+    }
+
     CHECK(connect_delayed == 1);
     CHECK(connected == 1);
     CHECK(total == expected_event_count);
--- a/tests/monitor.cpp
+++ b/tests/monitor.cpp
@@ -124,12 +124,36 @@ TEST_CASE("monitor init get event count", "[monitor]")
     CHECK_FALSE(monitor.get_event(eventMsg, address, zmq::recv_flags::dontwait));
     s.init();
 
-    while (total < expected_event_count)
+    SECTION("get_event")
     {
-        if (!monitor.get_event(eventMsg, address))
-            continue;
+        while (total < expected_event_count)
+        {
+            if (!monitor.get_event(eventMsg, address))
+                continue;
 
-        lbd_count_event(eventMsg);
+            lbd_count_event(eventMsg);
+        }
+
+    }
+
+    SECTION("poll get_event")
+    {
+        while (total < expected_event_count)
+        {
+            zmq::pollitem_t items[] = {
+                { monitor.handle(), 0, ZMQ_POLLIN, 0 },
+            };
+
+            zmq::poll(&items[0], 1, 100);
+
+            if (!(items[0].revents & ZMQ_POLLIN)) {
+                continue;
+            }
+
+            CHECK(monitor.get_event(eventMsg, address));
+
+            lbd_count_event(eventMsg);
+        }
     }
 
     CHECK(connect_delayed == 1);
--- a/tests/monitor.cpp
+++ b/tests/monitor.cpp
@@ -74,7 +74,7 @@ TEST_CASE("monitor move assign", "[monitor]")
     }
 }
 
-TEST_CASE("monitor init event count", "[monitor]")
+TEST_CASE("monitor init check event count", "[monitor]")
 {
     common_server_client_setup s{false};
     mock_monitor_t monitor;
@@ -90,6 +90,51 @@ TEST_CASE("monitor init event count", "[monitor]")
     CHECK(monitor.connect_delayed == 1);
     CHECK(monitor.connected == 1);
     CHECK(monitor.total == expected_event_count);
+}
+
+TEST_CASE("monitor init get event count", "[monitor]")
+{
+    common_server_client_setup s{ false };
+    zmq::monitor_t monitor;
+
+    const int expected_event_count = 2;
+    monitor.init(s.client, "inproc://foo");
+
+    int total{ 0 };
+    int connect_delayed{ 0 };
+    int connected{ 0 };
+
+    auto lbd_count_event = [&](const zmq_event_t& event) {
+        switch (event.event)
+        {
+        case ZMQ_EVENT_CONNECT_DELAYED:
+            connect_delayed++;
+            total++;
+            break;
+
+        case ZMQ_EVENT_CONNECTED:
+            connected++;
+            total++;
+            break;
+        }
+    };
+
+    zmq_event_t eventMsg;
+    std::string address;
+    CHECK_FALSE(monitor.get_event(eventMsg, address, zmq::recv_flags::dontwait));
+    s.init();
+
+    while (total < expected_event_count)
+    {
+        if (!monitor.get_event(eventMsg, address))
+            continue;
+
+        lbd_count_event(eventMsg);
+    }
+
+    CHECK(connect_delayed == 1);
+    CHECK(connected == 1);
+    CHECK(total == expected_event_count);
 }
 
 TEST_CASE("monitor init abort", "[monitor]")
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -2414,6 +2414,15 @@ class monitor_t
         _socket = socket_ref();
     }
 #endif
+
+    void close() ZMQ_NOTHROW
+    {
+#ifdef ZMQ_EVENT_MONITOR_STOPPED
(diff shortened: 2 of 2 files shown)


lp35:
Additional information:

- get_event() is not compatible with zmq version < 4
- if legacy compatibility with 3.x can be dropped for check_event(), the new get_event() method can be used to avoid duplicated code in check_event()

zmq.hpp
```diff
@@ -2224,6 +2224,53 @@ class monitor_t
         on_monitor_started();
     }
 
+#if ZMQ_VERSION_MAJOR >= 4
+    bool get_event(zmq_event_t& eventMsg, std::string& address, zmq::recv_flags flags = zmq::recv_flags::none)
```
sigiesec, line 2238: I am not sure about the signature of this method. What about 
```
std::optional<std::pair<zmq_event_t, std::string>> get_event(zmq::recv_flags flags = zmq::recv_flags::none)
```
(This requires C++17 this way, but we can also do this with the fallback to `detail::trivial_optional` as done for other types)

This removes the ambiguity of whether the original value of `eventMsg` and `address` are used and in what cases they are modified.

zmq.hpp
```diff
@@ -2224,6 +2224,53 @@ class monitor_t
...
+
+        {
+            message_t msg;
+            int rc = zmq_msg_recv(msg.handle(), _monitor_socket.handle(),
+                static_cast<int>(flags));
+
+            if (rc == -1)
+            {
+                if (zmq_errno() == ETERM || zmq_errno() == EAGAIN)
+                    return false;
```
sigiesec, line 2254: Why does this also returns false on `ETERM`? There's no way to distinguish `EAGAIN` and `ETERM` this way, and this seems inconsistent with the behavior of `zmq::socket_t::recv`.

zmq.hpp
```diff
@@ -2224,6 +2224,53 @@ class monitor_t
...
+            memcpy(&eventMsg.value, data, sizeof(int32_t));
+        }
+
+        message_t addrMsg;
+        int rc = zmq_msg_recv(addrMsg.handle(), _monitor_socket.handle(),
+            static_cast<int>(flags));
+
+        if (rc == -1)
+        {
+            if (zmq_errno() == ETERM)
```
sigiesec, line 2271: Again, why does this return `false` rather than throwing?

zmq.hpp
```diff
@@ -2224,6 +2224,53 @@ class monitor_t
...
+    {
+        assert(_monitor_socket);
+
+        eventMsg.event = 0;
+        eventMsg.value = 0;
+        address = std::string();
+
+        {
+            message_t msg;
+            int rc = zmq_msg_recv(msg.handle(), _monitor_socket.handle(),
```
sigiesec, line 2248: Can't this use `_monitor_socket.recv`?

zmq.hpp
```diff
@@ -2224,6 +2224,53 @@ class monitor_t
...
+            }
+
+            const char *data = msg.data<char>();
+            memcpy(&eventMsg.event, data, sizeof(uint16_t));
+            data += sizeof(uint16_t);
+            memcpy(&eventMsg.value, data, sizeof(int32_t));
+        }
+
+        message_t addrMsg;
+        int rc = zmq_msg_recv(addrMsg.handle(), _monitor_socket.handle(),
```
sigiesec, line 2266: Like above, can't this use `_monitor_socket.recv`?

tests/monitor.cpp
```diff
@@ -92,6 +92,51 @@ TEST_CASE("monitor init event count", "[monitor]")
...
+    zmq::monitor_t monitor;
+
+    const int expected_event_count = 2;
+    monitor.init(s.client, "inproc://foo");
+
+    int total{ 0 };
+    int connect_delayed{ 0 };
+    int connected{ 0 };
+
+    auto lbd_count_event = [&](const zmq_event_t& event) {
```
sigiesec, line 107: What does `lbd` stand for?

tests/monitor.cpp
```diff
@@ -124,12 +124,36 @@ TEST_CASE("monitor init get event count", "[monitor]")
...
+
+    SECTION("poll get_event")
+    {
+        while (total < expected_event_count)
+        {
+            zmq::pollitem_t items[] = {
+                { monitor.handle(), 0, ZMQ_POLLIN, 0 },
+            };
+
+            zmq::poll(&items[0], 1, 100);
```
sigiesec, line 147: Please use the overload taking `std::chrono::milliseconds`.

zmq.hpp
```diff
@@ -2224,6 +2224,14 @@ class monitor_t
         on_monitor_started();
     }
 
+    operator void *() ZMQ_NOTHROW { return handle(); }
+
+    operator void const *() const ZMQ_NOTHROW { return handle(); }
```
sigiesec, line 2229: I'd rather not add these operators. `socket_t` and `context_t` have them because they are direct correspondents of the underlying C data structures. But that's not the case here, and it's not obvious that a cast to `void*` returns the monitor socket handle.

zmq.hpp
```diff
@@ -2224,6 +2224,14 @@ class monitor_t
...
+
+    ZMQ_NODISCARD void *handle() ZMQ_NOTHROW { return _monitor_socket.handle(); }
+
+    ZMQ_NODISCARD const void *handle() const ZMQ_NOTHROW { return _monitor_socket.handle(); }
```
sigiesec, line 2233: Similar to the comment above, I'd use more specific names for these functions. We are in `monitor_t`, so `monitor` is probably redundant, but i'd use `socket_handle`.

Also, the `const` overload should be removed. There are no functions in the C zmq API that accept a `const void*` socket. I now they exist for other classes, but that's rather a legacy.

zmq.hpp
```diff
@@ -2232,6 +2232,8 @@ class monitor_t
 
     ZMQ_NODISCARD const void *handle() const ZMQ_NOTHROW { return _monitor_socket.handle(); }
 
+    operator socket_ref() ZMQ_NOTHROW { return (zmq::socket_ref) _monitor_socket; }
```
sigiesec, line 2235: Again, I'd not add an operator here but rather a named `socket()` member function. Actually, then the `handle`/`socket_handle` function is not necessary at all, since its handle can be accessed through the `socket_ref`.

zmq.hpp
```diff
@@ -2414,6 +2414,15 @@ class monitor_t
         _socket = socket_ref();
     }
 #endif
+
+    void close() ZMQ_NOTHROW
```
sigiesec, line 2418: Please also add a test for this.

zmq.hpp
lp35: Ok, I'm not very familiar with std::optional and will try to implement this way

zmq.hpp
lp35: Check https://github.com/zeromq/cppzmq/blob/master/zmq.hpp#L2274.

I just kept same behaviour implemented in check_event(). If this is not mandatory, I can just throw an exception and not use  a std::optionnal as described above, but just std::pair<>

zmq.hpp
lp35: See my comment above

zmq.hpp
lp35: Of course, but then it is not possible to use the same behaviour as the old check_event

tests/monitor.cpp
lp35: "lbd" is short for "lambda". I like to differentiate "callback" methods over classical variable in my code.

tests/monitor.cpp
lp35: Roger that

zmq.hpp
lp35: Agreed.

zmq.hpp
lp35: Agreed.

zmq.hpp
lp35: Agreed.

zmq.hpp
lp35: Will be done.

zmq.hpp
lp35: See prev comment

> Thanks for the extensions, I think these will be really useful.
> 
> Please see my comments on individual issues.
> 
> Also, please remember to reformat all commits with `clang-format`.
> 

Could you add this to the contribution guidelines of the project?

> Most of the travis builds are failing right now, could you give them a look and fix them, please?

I don't like to put effort on something that will be not merged. I was waiting for a green thumb before putting more work on this, as I did not asked anyone about the design of this PR ;)

Thanks for reviewing, will try to solve all of those next week. BR

zmq.hpp
sigiesec: Hm, yes, that's right. It's inconsistent one way or the other, but I think the behaviour of `zmq::socket_t::recv` is the appropriate one.

tests/monitor.cpp
sigiesec: Hm. I am not sure if it's necessary to use a prefix at all (if it were a function pointer, would it make a difference?), but if you feel it's good to have, please use `lambda_`.

sigiesec:
Any update on this? Is there anything you need to address the comments? Sorry I didn't get back until now, but from reading your responses I assumed that most things are clear and undisputed.

lp35:
Hi,

Thanks for pinging back. I'm struggling to find some time to fix those issues. I will try to do my best in the next couple of weeks.

Cheers

zmq.hpp
lp35: what we should implement in this case? Maybe we shall throw an exception?

PR #416, opened 2020-05-16
Problem: Feature checking inconsistencies

gummif:
Solution: Unify string_view feature checks. Document how users are expected to use exported macros and how they can define macros to control behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,15 @@ int main()
 }
 ```
 
+Compatibility Guidelines
+========================
+
+The users of cppzmq are expected to follow the guidelines below to ensure not to break when upgrading cppzmq to newer versions (non-exhaustive list):
+
+* Do not depend on any macros defined in cppzmq unless explicitly declared public here.
+
+The following macros may be used by consumers of cppzmq: `CPPZMQ_VERSION`, `CPPZMQ_VERSION_MAJOR`, `CPPZMQ_VERSION_MINOR`, `CPPZMQ_VERSION_PATCH`.
+
 Contribution policy
 ===================
 
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -103,18 +103,29 @@
 #include <tuple>
 #include <memory>
 #endif
-#ifdef ZMQ_CPP17
-#ifdef __has_include
-#if __has_include(<optional>)
-#include <optional>
-#define ZMQ_HAS_OPTIONAL 1
-#endif
-#if __has_include(<string_view>)
-#include <string_view>
-#define ZMQ_HAS_STRING_VIEW 1
-#endif
+
+#if defined(__has_include) && defined(ZMQ_CPP17)
+#define CPPZMQ_HAS_INCLUDE_CPP17(X) __has_include(X)
+#else
+#define CPPZMQ_HAS_INCLUDE_CPP17(X) 0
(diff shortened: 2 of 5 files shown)


gummif:
I am not sure if the CPPZMQ_HAS_* macros need to be customizable and documented. Maybe I should remove that from the readme?

zmq.hpp
```diff
@@ -103,18 +103,28 @@
 #include <tuple>
 #include <memory>
 #endif
+
 #ifdef ZMQ_CPP17
 #ifdef __has_include
 #if __has_include(<optional>)
-#include <optional>
-#define ZMQ_HAS_OPTIONAL 1
+#define CPPZMQ_HAS_OPTIONAL 1
```
sigiesec, line 110: If you want to make this overridable by a user, this needs to be guarded by an `#ifndef CPPZMQ_HAS_OPTIONAL`.

zmq.hpp
```diff
@@ -103,18 +103,28 @@
...
 #endif
 #if __has_include(<string_view>)
-#include <string_view>
-#define ZMQ_HAS_STRING_VIEW 1
+#define CPPZMQ_HAS_STRING_VIEW 1
```
sigiesec, line 113: Similarly, if you want to make this overridable by a user, this needs to be guarded by an `#ifndef CPPZMQ_HAS_STRING_VIEW`.

README.md
```diff
@@ -84,6 +84,18 @@ int main()
 }
 ```
 
+Compatibility Guidelines
+========================
+
+The users if cppzmq are expected follow the guidelines below for breakage-free behaviour when upgrading cppzmq to newer versions (non-exhaustive list):
+
+* Do not depend on any macros defined in cppzmq unless explicitly exported.
```
sigiesec, line 92: I would use a different term than `exported` here. Under `exported`, I'd expect some technical mechanism, but you simply mean that it is declared as such in the readme below.

What about saying:
```suggestion
* Do not depend on any macros defined in cppzmq unless explicitly declared public here.
```

README.md
```diff
@@ -84,6 +84,18 @@ int main()
...
+
+The following macros are exported: `CPPZMQ_VERSION`, `CPPZMQ_VERSION_MAJOR`, `CPPZMQ_VERSION_MINOR`, `CPPZMQ_VERSION_PATCH`. The following macros may be defined to opt in/out of behaviour:
```
sigiesec, line 94: These are two distinct aspects, and should be separated better. One is lead to think that the second sentence is related to exported/public macros as well, but on the contrary, it is related to macros that are imported/consumed. This should be made clearer.

README.md
```diff
@@ -84,6 +84,18 @@ int main()
 }
 ```
 
+Compatibility Guidelines
+========================
+
+The users if cppzmq are expected follow the guidelines below for breakage-free behaviour when upgrading cppzmq to newer versions (non-exhaustive list):
```
sigiesec, line 90: Some editorial improvements:
```suggestion
The users of cppzmq are expected to follow the guidelines below to ensure not to break when upgrading cppzmq to newer versions (non-exhaustive list):
```

sigiesec:
Thanks for addressing the comments, and sorry it took so long to merge it!

PR #418, opened 2020-05-16
Problem: buffer definitions after message_t

gummif:
Soluion: move buffer definitions to a more logic location

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -242,7 +242,314 @@ struct is_range<
 };
 
 } // namespace detail
+
+// mutable_buffer, const_buffer and buffer are based on
+// the Networking TS specification, draft:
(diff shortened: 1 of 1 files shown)


gummif:
This was part of a previous change that added constructors to message_t that take buffer related to deprecating another ctor. Yes, the diff looks bad, but it is just a simple move. Should I try to split this into smaller commits that might look better individually?

sigiesec:
I think this shouldn't be done. Splitting it up into smaller commits might be slightly better, but doesn't change the fundamental problem that this breaks `blame`ing without a significant benefit. We might do a big re-haul of the file at some point of time, but then the structure of the whole file should be rethought.

gummif:
OK I will close this for now and work on the other PR without using the zmq::buffer overloads, since those are not really required for that change.

PR #423, opened 2020-05-26
Add an example

serg06:
I had a really tough time figuring out how to do simple multi-threaded pub-sub with cppzmq; I finally figured it out so I'd like to add it.

### DIFF
--- a/examples/pubsub_multithread_inproc.cpp
+++ b/examples/pubsub_multithread_inproc.cpp
@@ -98,6 +98,11 @@ int main()
 	zmq::context_t ctx;
 
 	auto thread1 = std::async(std::launch::async, PublisherThread, &ctx);
+	
+	// Give publisher time to bind
+	// "the server must issue a bind before any client issues a connect"
+	Sleep(10);
+
 	auto thread2 = std::async(std::launch::async, SubscriberThread1, &ctx);
 	auto thread3 = std::async(std::launch::async, SubscriberThread2, &ctx);
 	thread1.wait();
--- a/examples/pubsub_multithread_inproc.cpp
+++ b/examples/pubsub_multithread_inproc.cpp
@@ -5,95 +5,95 @@
 #include "zmq.hpp"
 #include "zmq_addon.hpp"
 
-void PublisherThread(zmq::context_t* ctx)
-{
-	//  Prepare publisher
-	zmq::socket_t publisher(*ctx, zmq::socket_type::pub);
-	publisher.bind("inproc://#1");
+void PublisherThread(zmq::context_t *ctx) {
+    //  Prepare publisher
+    zmq::socket_t publisher(*ctx, zmq::socket_type::pub);
+    publisher.bind("inproc://#1");
 
-	// Give the subscribers a chance to connect, so they don't lose any messages
-	std::this_thread::sleep_for(std::chrono::milliseconds(20));
+    // Give the subscribers a chance to connect, so they don't lose any messages
+    std::this_thread::sleep_for(std::chrono::milliseconds(20));
 
-	while (true) {
-		//  Write three messages, each with an envelope and content
-		publisher.send(zmq::str_buffer("A"), zmq::send_flags::sndmore);
-		publisher.send(zmq::str_buffer("Message in A envelope"));
-		publisher.send(zmq::str_buffer("B"), zmq::send_flags::sndmore);
-		publisher.send(zmq::str_buffer("Message in B envelope"));
-		publisher.send(zmq::str_buffer("C"), zmq::send_flags::sndmore);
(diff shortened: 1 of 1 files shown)


examples/pubsub_multithread_inproc.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+int main()
+{
+	/*
+	 * No I/O threads are involved in passing messages using the inproc transport.
+	 * Therefore, if you are using a ØMQ context for in-process messaging only you
+	 * can initialise the context with zero I/O threads.
+	 *
+	 * Source: http://api.zeromq.org/2-1:zmq-inproc
```
gummif, line 83: Maybe link to a more recent version http://api.zeromq.org/4-3:_start

examples/pubsub_multithread_inproc.cpp
```diff
@@ -0,0 +1,114 @@
...
+	 *     Thread2: [A] Message in A envelope
+	 *     Thread2: [B] Message in B envelope
+	 *     Thread3: [A] Message in A envelope
+	 *     Thread3: [B] Message in B envelope
+	 *     Thread3: [C] Message in C envelope
+	 */
+}
+
+#ifdef _WIN32
+int CALLBACK WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine, int nCmdShow)
```
gummif, line 110: is this really required?

examples/pubsub_multithread_inproc.cpp
```diff
@@ -0,0 +1,114 @@
...
+		zmq::message_t address;
+		zmq::recv_result_t result = subscriber.recv(address);
+
+		//  Read message contents
+		zmq::message_t contents;
+		result = subscriber.recv(contents);
+
+		std::stringstream out;
+		out << "Thread3: [" << address.to_string_view() << "] " << contents.to_string_view() << std::endl;
+		OutputDebugString(out.str().c_str());
```
gummif, line 72: Use `std::cout` directly instead.

examples/pubsub_multithread_inproc.cpp
```diff
@@ -0,0 +1,114 @@
...
+	//  Prepare our context and subscriber
+	zmq::socket_t subscriber(*ctx, zmq::socket_type::sub);
+	subscriber.connect("inproc://#1");
+
+	//  Thread3 opens ALL envelopes
+	subscriber.setsockopt(ZMQ_SUBSCRIBE, "", 0);
+
+	while (1) {
+		//  Read envelope address
+		zmq::message_t address;
```
gummif, line 63: A better name would be `topic` in my opinion

examples/pubsub_multithread_inproc.cpp
```diff
@@ -0,0 +1,114 @@
...
+		zmq::recv_result_t result = subscriber.recv(address);
```
gummif, line 64: I think there should be some error checking, at least an `assert`. You could also use `zmq::recv_multipart` to simplify the message handling.

examples/pubsub_multithread_inproc.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+void SubscriberThread2(zmq::context_t* ctx)
+{
+	//  Prepare our context and subscriber
+	zmq::socket_t subscriber(*ctx, zmq::socket_type::sub);
+	subscriber.connect("inproc://#1");
+
+	//  Thread3 opens ALL envelopes
+	subscriber.setsockopt(ZMQ_SUBSCRIBE, "", 0);
```
gummif, line 59: The subscribes can use the type safe way `subscriber.set(zmq::sockopt::subscribe, "");`

examples/pubsub_multithread_inproc.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+void SubscriberThread1(zmq::context_t* ctx)
+{
+	//  Prepare subscriber
+	zmq::socket_t subscriber(*ctx, zmq::socket_type::sub);
+	subscriber.connect("inproc://#1");
+
+	//  Thread2 opens "A" and "B" envelopes
+	subscriber.setsockopt(zmq::sockopt::subscribe, "A", 1);
+	subscriber.setsockopt(zmq::sockopt::subscribe, "B", 1);
```
gummif, line 37: I think that should be `subscriber.set(zmq::sockopt::subscribe, "B");`

serg06: Oh my bad, I misread the first time. Fixed.

sigiesec:
Thanks for adding the examples, that's really helpful! Could you just reformat that using clang-format?

sigiesec:
Thanks! Sorry for taking so long to merge this.

PR #424, opened 2020-06-15
Issue: cannot use cppzmq as cmake subproject

hshakula:
The easiest way to accomplish this is to skip `find_package`s when `libzmq` or `libzmq-static` targets exist.


hshakula:
I've missed that these changes already in master, closing.

Issue #426, opened 2020-06-27
Bug: Sometimes ZMQ_CPP17 not defined on MSVC


serg06:
### System:

- Windows 10
- VS 2019 16.6.2
- MSVC
- Using latest C++20 preview

### Issue

Occasionally `ZMQ_CPP17` doesn't get defined so features like `zmq::message_t::to_string_view()` don't work.

### Explanation

This is where `ZMQ_CPP17` is defined:

```
#if (defined(__cplusplus) && __cplusplus >= 201703L) || (defined(_HAS_CXX17) && _HAS_CXX17 == 1)
    #define ZMQ_CPP17
#endif
```

The first check `(defined(__cplusplus) && __cplusplus >= 201703L)` fails because VS doesn't give `__cplusplus` a proper value if you don't set the [/Zc:__cplusplus](https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=vs-2019) compiler flag.

The second check `(defined(_HAS_CXX17) && _HAS_CXX17 == 1)` fails **if the user hasn't included `vcruntime.h` yet**.

### Potential solutions

1. Before the _HAS_CXX17 check, if compiling with MSVC, include `vcruntime.h`.
2. Move `#include <zmq.h>` above the check because `zmq.h` includes `vcruntime.h`.
3. Move all the includes before the check - [that's how nlohmann/json does it](https://github.com/nlohmann/json/blob/master/include/nlohmann/detail/macro_scope.hpp).

### Temporary workarounds

If anyone runs into this issue, just add the [/Zc:__cplusplus](https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=vs-2019) compiler flag.

gummif:
Thanks for that. After some investigation I think we may want to check _MSVC_LANG instead of the _HAS_CXX macros.

zyede:
How did you compile this library for vs2019?，I tried every method but failed to compile

gummif:
If you are having problems you can submit an issue with more details what fails.

serg06:
> How did you compile this library for vs2019?，I tried every method but failed to compile

I did it by adding `#include <vcruntime.h>` above `#include <zmq.h>`.

PR #429, opened 2020-07-11
Problem: C++ version not correctly estimated

gummif:
Solution: Include headers before checking macros and add additional checks for _MSVC_LANG.

Fixes https://github.com/zeromq/cppzmq/issues/426

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -92,6 +92,9 @@
 #define ZMQ_CONSTEXPR_VAR const
 #define ZMQ_CPP11_DEPRECATED(msg)
 #endif
+#if defined(ZMQ_CPP14) && (!defined(_MSC_VER) || _MSC_VER > 1900)
+#define ZMQ_EXTENDED_CONSTEXPR
+#endif
 #if defined(ZMQ_CPP17)
 #define ZMQ_INLINE_VAR inline
 #else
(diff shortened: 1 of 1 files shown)


zmq.hpp
```diff
@@ -32,20 +32,34 @@
 #endif
 #endif
 
+// included here for _HAS_CXX* macros
+#include <zmq.h>
+
+#if defined(_MSCV_LANG)
```
sigiesec, line 38: I guess this should be `_MSVC_LANG`? Also in the line below.

zmq.hpp
```diff
@@ -32,20 +32,34 @@
...
+#define CPPZMQ_LANG _MSCV_LANG
+#elif defined(__cplusplus)
+#define CPPZMQ_LANG __cplusplus
+#else
+#define CPPZMQ_LANG 0
```
sigiesec, line 43: I think this should rather be some `#error`. If `__cplusplus` is not defined, we are not compiling in C++ mode, and that can't work.

sigiesec (changes requested):
Sorry for taking so long to look at this.

Is there a way we could test this is working as expected? In particular with MSVC? We can merge this independently of addressing that, but it would be generally nice, as it might easily break. (Obviously it doesn't work as of now as the define used by MSVC is misspelled, still all jobs seem fine)

zmq.hpp
gummif: Yes, good catch.

gummif:
This breaks VS 2015 for some reason. Will take a look later.

PR #432, opened 2020-07-18
Bugfix: socket_t move assignment doesn't initialize ctxptr

or17191:
Until now, we only assigned the _handle on
`zmq::socket_t::operator=(socket_t&&)`. This manifests when trying to
monitor a socket initialized by that constructor.

To avoid changing the public interface of the socket_t class, we tested
for the specific monitor usecase, since it's the only class accessing
zmq::socke_t::ctxptr.

NOTE: When running the new unit-test without the fix, it might hang on
`zmq_socket_monitor(socket_, NULL, 0)`. We haven't figured out the cause
for that, but we deemed it unimportant (or at least out of scope).

### DIFF
--- a/tests/monitor.cpp
+++ b/tests/monitor.cpp
@@ -142,4 +142,19 @@ TEST_CASE("monitor init abort", "[monitor]")
     monitor.abort();
     thread.join();
 }
+
+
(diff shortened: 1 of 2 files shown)


sigiesec:
Thanks for the fix, and for adding a test in particular! And sorry for keeping you waiting for so long!

PR #450, opened 2020-10-10
Problem: Examples are not compiled

gummif:
Solution: Compile the examples when tests are compiled and using C++11 or greater.


sigiesec:
Thanks! That's a really good idea. Having non-working examples is very annoying, and ensuring they build is definitely worthwhile.

PR #451, opened 2020-10-10
Problem: Poller size function missing

gummif:
Solution: Add it to poller_t

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -2653,6 +2653,15 @@ template<typename T = no_user_data> class poller_t
         throw error_t();
     }
 
+#if ZMQ_VERSION >= ZMQ_MAKE_VERSION(4, 3, 3)
+    size_t size() const noexcept
+    {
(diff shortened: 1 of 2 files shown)


zmq.hpp
```diff
@@ -2653,6 +2653,15 @@ template<typename T = no_user_data> class poller_t
         throw error_t();
     }
 
+#if ZMQ_VERSION >= ZMQ_MAKE_VERSION(4, 3, 3)
+    size_t size() const noexcept
+    {
+        int rc = zmq_poller_size(const_cast<void *>(poller_ptr.get()));
+        ZMQ_ASSERT(rc >= 0);
+        return static_cast<size_t>(std::max(rc, 0));
```
sigiesec, line 2661: `ZMQ_ASSERT` always crashes on failure. The `std::max` can therefore be removed.

sigiesec: Ah no, I confused that with libzmq's zmq_assert...

sigiesec:
Thanks!

PR #452, opened 2020-10-10
Problem: zmq_fd_t not used/available in zmq

gummif:
Solution: Make zmq::fd_t available and use in poller_event

This is currently available as zmq::sockopt::cppzmq_fd_t, so this is techincally a breaking change. I could add a typedef there for backward compatibility reasons if requested.

### DIFF
--- a/tests/socket.cpp
+++ b/tests/socket.cpp
@@ -270,8 +270,7 @@ TEST_CASE("socket check integral options", "[socket]")
     check_integral_opt_get<int>(zmq::sockopt::events, router, "events");
 #endif
 #ifdef ZMQ_FD
-    check_integral_opt_get<zmq::sockopt::cppzmq_fd_t>(zmq::sockopt::fd, router,
-                                                      "fd");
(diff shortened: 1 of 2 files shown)


sigiesec:
> This is currently available as zmq::sockopt::cppzmq_fd_t, so this is techincally a breaking change. I could add a typedef there for backward compatibility reasons if requested.

Yes, please add a typedef. It's easy to avoid breakage that way, so even if it's unlikely, we should do that. You might add a comment that using `zmq::fd_t` is preferred. (I think ZMQ_DEPRECATED doesn't work there?)

sigiesec:
This is still breaking the Windows builds at
```
  C:\projects\cppzmq\zmq.hpp(1398): error C2062: type 'int' unexpected [C:\projects\cppzmq\Build\tests\unit_tests.vcxproj]
```

Could you give that another look?

gummif:
Finally took a look at this, don't know if there is a good way to do this so just used a comment.

PR #459, opened 2020-10-25
Problem: conversion warnings

gummif:
Solution: Silence them with casts


sigiesec:
Can you rebase the PR please, so we can see the CI working after the catch fix?

PR #460, opened 2020-10-25
Problem: if constexpr warnings on MSVC

gummif:
Solution: Use if constexpr where possible in C++17

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -97,8 +97,10 @@
 #endif
 #if defined(ZMQ_CPP17)
 #define ZMQ_INLINE_VAR inline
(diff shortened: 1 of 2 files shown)


sigiesec:
It's not clear to me which specific warnings this fixes, but while this adds some complexity that doesn't seem strictly necessary, I am fine with this.

Can you rebase the PR please, so we can see the CI working after the catch fix?

gummif:
If I remember correctly then I get this warning where I use the library https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4127?view=msvc-160 . I will rebase it.

PR #482, opened 2021-03-06
Fix for potential memory leak in monitor_t::check_event

kevle:
The objects _eventMsg_ and _addrMsg_ have the type **zmq_msg_t**.
Since the event handlers might throw, this could have caused a leak of data associated with _eventMsg_.
For _addrMsg_, the constructor of std::string could have thrown ( although this is unlikely ).

This PR replaces raw  **zmq_msg_t** objects with **zmq::message_t** objects in _monitor_t::check_event_.

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -2309,8 +2309,7 @@ class monitor_t
     {
         assert(_monitor_socket);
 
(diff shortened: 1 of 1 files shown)


gummif:
Looks good to me.

Issue #448, opened 2020-10-05
zmq::multipart_t::send() only takes socket_t & argument, not socket_ref


jeroenmuller:
Using 4.8.0 with draft APIs enabled. 

I am using the poller with zmq::multipart_t. The poller event gives me a socket_ref. However, multipart_t::send() only accepts a reference to a socket (not the socket_ref class). Therefore, I get a compiler error of this form:

```
src/mesh.cpp:36:29: error: cannot convert ‘zmq::socket_ref’ to ‘zmq::socket_t&’
```

An example of the code that leads to this issue (but I assume every case where we want to pass a socket_ref to multipart_t would lead to the same issue):

```cpp
std::vector<zmq::poller_event<size_t>> events{10};
size_t num_events;
while ((num_events = d_in_poller.wait_all(events, std::chrono::milliseconds{0}))) {
	for (size_t i = 0; i != num_events; i++) {
		zmq::multipart_t multipart;
		multipart.recv(*(events[i].socket);
```

gummif:
Good point. This can be easily fixed, you are welcome to submit a PR if you can.

gummif:
An alternative is to use the `recv_multipart` and `send_multipart` functions.

PR #487, opened 2021-04-17
Add support for socket_ref to multipart_t ctor/send/recv 

barometz:
This adds socket_ref support to `multipart_t`'s constructor and `send`/`recv` member functions, resolving issue #448 . All previously took a `socket_t&` instead. I don't *think* this will break anything, but if preferred I can add alias functions for `socket_t&` back in.

### DIFF
--- a/tests/multipart.cpp
+++ b/tests/multipart.cpp
@@ -1,6 +1,10 @@
 #include <catch.hpp>
 #include <zmq_addon.hpp>
 
+static_assert(std::is_same<decltype(zmq::multipart_t().send(zmq::socket_ref())), bool>::value);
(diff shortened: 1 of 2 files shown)


barometz:
The remaining test error is `The command "pip install --user cpp-coveralls" failed and exited with 1 during .`, which I don't think I can take credit for.

gummif (approved):

sigiesec:
> This adds socket_ref support to `multipart_t`'s constructor and `send`/`recv` member functions, resolving issue #448 . All previously took a `socket_t&` instead.
> I don't _think_ this will break anything, but if preferred I can add alias functions for `socket_t&` back in.

Given the implicit `socket_t::operator socket_ref` I think it's fine to just change it. Thanks for the fix!

Issue #462, opened 2020-11-03
Missing dgram in zmq::socket_type


CelticMinstrel:
Self-explanatory. Even when building with draft APIs, no dgram option is included in the enum.

gummif:
A PR is welcome, but until it is added it's possible to simply static_cast the value to the enum type.

PR #488, opened 2021-04-17
Add all draft socket types up to libzmq 4.3.4

barometz:
Resolving issue #462 and then some.

This adds the GATHER, SCATTER, CHANNEL, PEER and DGRAM socket types to `zmq::socket_type`. I can't find *any* documentation about DGRAM but since it was requested and does exist in zmq.h I figured I'd include it. I'm guessing that's when you map ZeroMQ messages to UDP datagrams?

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -2030,6 +2030,13 @@ enum class socket_type : int
     client = ZMQ_CLIENT,
     radio = ZMQ_RADIO,
     dish = ZMQ_DISH,
+    gather = ZMQ_GATHER,
+    scatter = ZMQ_SCATTER,
+    dgram = ZMQ_DGRAM,
+#endif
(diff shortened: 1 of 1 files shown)


zmq.hpp
```diff
@@ -2030,6 +2030,13 @@ enum class socket_type : int
     client = ZMQ_CLIENT,
     radio = ZMQ_RADIO,
     dish = ZMQ_DISH,
+    gather = ZMQ_GATHER,
```
gummif, line 2033: Should probably guard all with `#ifdef`s?

barometz: They're all behind `#if defined(ZMQ_BUILD_DRAFT_API) && ZMQ_VERSION >= ZMQ_MAKE_VERSION(4, 2, 0)` - I verified the version for all.

PR #496, opened 2021-06-27
Update CMake in CI

albestro:
This PR for updating CMake requirements in CI.

This is a prerequisite for #466.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -106,7 +106,15 @@ matrix:
       # packages:
         # - clang-5.0
 
+before_script:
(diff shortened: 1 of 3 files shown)


albestro:
I've updated just Travis and not AppVeyor, because the latter one already has a quite recent version of CMake (I've seen 3.16 in some log).

Let me know if you prefer to start keeping aligned the CMake versions between the two different CIs.

gummif:
Thanks, don't see a reason not to update cmake to a newer version since anyone should be able to do that easily

PR #497, opened 2021-06-29
Problem: poll default timeout is deprecated

gummif:
Solution: default to std::chrono::milliseconds


gummif:
@sigiesec @bluca What is best practice regarding PRs here, should I ping you for review?

PR #509, opened 2021-07-19
Fix handshake event don't need DRAFT API in libzmq v4.3.0 and above

hirokiht:
A rather straightforward fix for allowing access to handshake events in stable libzmq v4.3.0 which they were promoted as stable API.


gummif:
Thanks

Issue #361, opened 2019-11-02
Windows MSYS discovery of libzmq breaks due to library names


ferdnyc:
Post-#267, I find that the `FindZeroMQ.cmake` from this repo is still failing on Windows, when invoked in a CMake project using the "MSYS Makefiles" generator, with ZeroMQ 4.3.1 installed from the MYS-official `mingw32/mingw-w64-i686-zeromq` (32-bit) and `mingw64/mingw-w64-x86_64-zeromq` (64-bit) packages via `pacman`.

(This also appears to be a different issue from #270, where I believe libzmq was self-built by the reporter. I'm not building libzmq, just installing the standard MSYS packages that include a `libzmq.pc` file.)

A little background. Also, apologies for the length of the forthcoming coredump:

`zmq.hpp` is actually included in the MSYS packages, so technically `cppzmq` isn't needed on (our) Windows at all, and `find_package(cppzmq)` will end up being a `NOTFOUND`.

However, since ZeroMQ doesn't install CMake targets for itself, and since our project uses ZeroMQ in cross-platform builds on other systems where `cppzmq` _is_ useful (various Linux distros like Fedora, where `cppzmq` is a separate package and **not** bundled with `libzmq`), I wanted to replace our home-grown `FindZMQ.cmake` with the `FindZeroMQ.cmake` from `cppzmq`. My goal was to have ZeroMQ itself always be detected the same way, instead of getting configured twice in different ways when `cppzmq` is present. Plus, `FindZeroMQ.cmake` created `IMPORTED` targets, where our old find module was still stuck in the results-variables dark ages.

The switchover worked a treat, on Linux and MacOS. But on our Windows builders, it broke in really weird, subtle ways.

The issue, I believe, comes down to this note, from [the "Imported Libraries" section](https://cmake.org/cmake/help/v3.16/command/add_library.html#imported-libraries) of the CMake `add_library()` docs:
<blockquote>An `UNKNOWN` library type is typically only used in the implementation of Find Modules. It allows the path to an imported library (often found using the `find_library()` command) to be used without having to know what type of library it is. <b>This is especially useful on Windows where a static library and a DLL’s import library both have the same file extension.</b></blockquote>

That point is relevant to these sections of the `FindZeroMQ.cmake` code:
https://github.com/zeromq/cppzmq/blob/fdb2f1397104f68100ea5fa1e288f1a4d0caf9ed/libzmq-pkg-config/FindZeroMQ.cmake#L6-L9
[...]

PR #510, opened 2021-07-28
Fixes https://github.com/zeromq/cppzmq/issues/361, by incorperating F…

Cazadorro:
Fixes https://github.com/zeromq/cppzmq/issues/361, by incorperating Ferdnyc's changes in OpenShot/libopenshot, previously cppzmq was not looking for the correct names, and rejected valid static library names when looking for zmq on windows, this fix aims to include those other names. 

I've only tested this under zmq from VCPKG with the static target triplet, it would be good for someone to test this against non static target to make sure this doesn't break anything that worked before (I can't test that on my end).


gummif:
Does this work for static-only libzmq (is that an option even)?

gummif:
I will just merge this now, looks like everything works in the CI.

PR #515, opened 2021-09-16
ZMQ_DEPRECATED doesn't fallback for unrecognized compiler

ashish-17:
If you use a compiler outside of _MSC_VER and _GNUC__, then the if-else check doesn't provide a fallback for ZMQ_DEPRECATED macro which leads to a build failure.
Specifically in my case , I am using Oracle's SunPro 5.13.0.

My suggestions is to fallback to noop in case you don't recognize a compiler. Addressed #516


gummif:
Thanks for that.

Issue #522, opened 2021-10-15
Return type of `sock.get(zmq::sockopt::type)`


jasujm:
The new `sock.get(zmq::sockopt::option)` API does a good job in retrieving socket options in a typesafe manner, and is definitely an improvement over the old `sock.getsockopt()` API.

But I think that `sock.get(zmq::sockopt::type)` in particular works illogically. Currently, this won't compile:

```
$ cat test.cc 
#include <zmq.hpp>

#include <cassert>

int main()
{
    zmq::context_t ctx;
    zmq::socket_t sock(ctx, zmq::socket_type::push);
    assert(sock.get(zmq::sockopt::type) == zmq::socket_type::push);
    return 0;
}
$ g++ -c test.cc 
In file included from /usr/include/c++/10/cassert:44,
                 from test.cc:3:
test.cc: In function ‘int main()’:
test.cc:9:41: error: no match for ‘operator==’ (operand types are ‘int’ and ‘zmq::socket_type’)
    9 |     assert(sock.get(zmq::sockopt::type) == zmq::socket_type::push);
      |            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^~ ~~~~~~~~~~~~~~~~~~~~~~
      |                    |                                         |
      |                    int                                       zmq::socket_type
```

Would it make more sense that in C++11 and above `sock.get(zmq::sockopt::type)` would return enumerator from `zmq::socket_type` instead of an `int`? `int` is neither typesafe nor used to represent socket types elsewhere in the library. This change would break backward compatibility slightly, but in a way that I'm sure any user of `zmq::sockopt::type` would accept.

What do you think?

jasujm:
That sounds good :). I created a PR which would do that: https://github.com/zeromq/cppzmq/pull/523

PR #523, opened 2021-10-21
Problem: `sock.get(zmq::sockopt::type)` is not typesafe

jasujm:
Solution: Add `zmq::sockopt::socket_type` that is used to `get()` an enumerator from `zmq::socket_type` instead of plain `int`.

It makes code like this compile:

  #include <zmq.hpp>

  #include <cassert>

  int main()
  {
      zmq::context_t ctx;
      zmq::socket_t sock(ctx, zmq::socket_type::push);
      assert(sock.get(zmq::sockopt::socket_type) == zmq::socket_type::push);
      return 0;
  }

This patch solves https://github.com/zeromq/cppzmq/issues/522

### DIFF
--- a/tests/socket.cpp
+++ b/tests/socket.cpp
@@ -406,7 +406,10 @@ TEST_CASE("socket check integral options", "[socket]")
 #endif
 #ifdef ZMQ_TYPE
     check_integral_opt_get<int>(zmq::sockopt::type, router, "type");
-#endif
+#ifdef ZMQ_CPP11
+    check_integral_opt_get<zmq::socket_type>(zmq::sockopt::socket_type, router, "socket_type");
+#endif // ZMQ_CPP11
+#endif // ZMQ_TYPE
 
 #ifdef ZMQ_HAVE_VMCI
 #ifdef ZMQ_VMCI_BUFFER_SIZE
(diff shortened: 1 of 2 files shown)


XieJiSS:
Hi @gummif , I encountered this error when trying to `#include "zmq.hpp"`:

```cpp
// This is the macro expansion result
using socket_type_t = integral_option<16, socket_type, false>; inline constexpr socket_type_t socket_type {}
// The error
"zmq::sockopt::socket_type" is not a type name
```

According to my linter, this is caused by
https://github.com/zeromq/cppzmq/blob/a98fa4a91d868a3844e5456741d6782cc1a8d98b/zmq.hpp#L1652

My workaround is:

```cpp
// socket_type seems to be resolved to zmq::sockopt::socket_type instead of the enum class zmq::socket_type
ZMQ_DEFINE_INTEGRAL_OPT(ZMQ_TYPE, socket_type, zmq::socket_type);
```

I'm not very familiar with cpp, so I'm wondering whether this should be considered as a bug introduced by the PR, or an issue caused by possible misconfiguration?

My environment: `#define CPPZMQ_LANG _MSVC_LANG` which extends to `201703L` and `#define _MSC_VER 1929`

PR #533, opened 2021-12-30
zmq_recv zmq_send handle EINTR

lyzyung:
An exception was raised when I used GperfTools because of EINTR error


gummif:
I guess this would be a breaking change.  Would a workaround catching this error in your code not work for you?

lyzyung:
I can catch this exception, but I don’t want to catch it everywhere

PR #553, opened 2022-04-21
Suppress ZMQ_EXTENDED_CONSTEXPR on gcc with partial C++14 support

gin-ahirsch:
cf. https://gcc.gnu.org/projects/cxx-status.html#cxx14


gin-ahirsch:
Had to change from checking `__cpp_constexpr` to a specific gcc version. Looks like gcc <= 5.3 errors, but still defines `__cpp_constexpr` to `201304` so that's insufficient.

gummif:
Thanks

PR #559, opened 2022-06-15
Updating Catch2 v2 to v2.13.9

nandanvasudevan:
Fixes issue with MINSIGSTKSZ which is no longer a constant -> https://github.com/JuliaLang/julia/issues/39822

Issue was fixed in https://github.com/catchorg/Catch2/releases/tag/v2.13.5

Does not build in Ubuntu 22.04 with GCC 11.2.0 without this update.


gummif:
Thanks

Issue #477, opened 2021-02-11
ZMQ_BUILD_DRAFT_API is not added to compiler definitions for cppzmq target when ENABLE_DRAFTS is ON.


melton1968:
When I run `cmake` to configure `cppzmq` with the `-DENALBE_DRAFTS=ON` flag and then build and install the `cppzmq` target, the corresponding `cppzmqTargets.cmake` configuration file does not set the `-DZMQ_BUILD_DRAFT_API` compiler option as expected.

The code in `CMakeLists.txt` looks correct:

```
if (ENABLE_DRAFTS)
    COMPILE_DEFINITIONS(-DZMQ_BUILD_DRAFT_API)
    set (pkg_config_defines "-DZMQ_BUILD_DRAFT_API=1")
else (ENABLE_DRAFTS)
    set (pkg_config_defines "")
endif (ENABLE_DRAFTS)
```

If I move that code block to below where the `cppzmq` target is defined and change the `add_definitions` method to `target_compile_definitions`:

```
    target_compile_definitions(cppzmq INTERFACE ZMQ_BUILD_DRAFT_API)
```

then `ZMQ_BUILD_DRAFT_API` correctly gets defined as a compiler option in the installed configuration files.

I am not a `cmake` expert, so I am uncertain if there is something I am missing, but this seems to be some type of adverse interaction between `add_definitions` and interface targets.

I am using `cmake` 3.19.4.

I can offer a pull request with the fix if that is useful.

jlucke86:
@melton1968 the CMake seems to check for a .git folder in the root directory and if it exists enable building the draft api support. I ran x86 build and also arm cross compiled cppzmq 4.6.0 and libzmq 4.3.2 a few weeks ago. I’m pretty sure I didn’t need any special flags.

I support another target platform where I believe I used the compiler definitions as it was building from tar.gz sources rather than git clones. I will try to check that and let you know which flags I used, or the build system uses.

Once you get the libraries compiled you still need to define the enable draft api’s #define before including zmq.hpp in your application otherwise the draft api code will not be exposed.

datax1969:
I am using cppzmq in-source via CMake's FetchContent facilities and run into this problem as well.  Rather than having to remember to define ZMQ_BUILD_DRAFT_API at higher levels, I think it should be put here as a target_compile_definition.  The fix that gets me going is:

```
line 49-54:
foreach (target cppzmq cppzmq-static)
  add_library(${target} INTERFACE)
[...]

PR #565, opened 2022-07-29
Remove ENABLE_DRAFTS option

stephanlachnit:
This option added ZMQ_BUILD_DRAFT_API to the compile definitions. However, this should be propagated by libzmq (see https://github.com/zeromq/libzmq/pull/4323 and https://github.com/zeromq/cppzmq/issues/477).

Closes https://github.com/zeromq/cppzmq/issues/477.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,7 +132,6 @@ jobs:
       run: |
         cmake -H. -Bbuild ${{ matrix.platform}} ${{ matrix.coverage }} \
           -DCMAKE_BUILD_TYPE=${BUILDTYPE} \
-          -DENABLE_DRAFTS=${{ matrix.drafts }} \
           -DCMAKE_CXX_STANDARD=${{ matrix.cppstd }}
         cmake --build build --config ${BUILDTYPE} -j ${THREADS}
         echo "CPPZMQ=${PWD}/build" >> ${GITHUB_ENV}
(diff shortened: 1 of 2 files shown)


gummif:
This change will probably break something for someone using slightly older versions of libzmq.

stephanlachnit:
> This change will probably break something for someone using slightly older versions of libzmq.

Yes, I would recommend merging this change only after a new release of libzmq (the commit in question is not part of a release yet), since it is fairly reasonable to expect people not to take a new version of cppzmq and an old version libzmq.

stephanlachnit:
@gummif since libzmq v4.3.5 has been released with https://github.com/zeromq/libzmq/issues/4194 included, this can merged now

stephanlachnit:
I also updated the CI to use 4.3.5. to actually use this change.

gummif:
Sorry for the delay. I will see if I have time soon to do a review.

stephanlachnit:
@gummif friendly ping to check this :) This is quite a small change

gummif:
OK I think this looks good. In the unlikely scenario someone has problem with enabling drafts (using older version of libzmq built with drafts and new version of cppzmq) they just need to add `target_compile_definitions(${target} PUBLIC ZMQ_BUILD_DRAFT_API)` to their cmake project.

stephanlachnit:
@gummif any chance for a new release with this fix?

gummif:
Yes, will try to do it soon.

PR #569, opened 2022-09-09
add a parameter for multipart_t::send

caicaiking:
I need send message and recv message to do ACK, when I not recv the ack message I need to send it again.  I don't want to use clone, because need time to copy.

### DIFF
--- a/zmq_addon.hpp
+++ b/zmq_addon.hpp
@@ -413,7 +413,7 @@ class multipart_t
     }
 
     // Send multipart message to socket
-    bool send(socket_ref socket, int flags = 0)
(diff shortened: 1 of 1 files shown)


caicaiking:
> Thos function is destructive so clear is a no-op it looks like. Copying messages is really the only way.

So, Can I add this control parameter for send funciton?

gummif:
What you really need to do is something like this
```c++
std::vector<zmq::const_buffer> bufs;
for (const auto& m : msgs)
  bufs.emplace_back(m.data(), m.size());
while (not done)
{
  auto ret = zmq::send_multipart(bufs);
  // repeat send until something
}
```
Make a vector/array of buffers and use the send_multipart function (which will not modify the messages).

PR #581, opened 2022-10-14
Fixes warning noexcept

Teebonne:
This type of function should not throw, adding noexcept

### DIFF
--- a/zmq_addon.hpp
+++ b/zmq_addon.hpp
@@ -343,10 +343,10 @@ class multipart_t
     multipart_t(message_t &&message) { add(std::move(message)); }
 
     // Move constructor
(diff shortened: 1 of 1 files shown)


gummif:
You must use the ZMQ_NOTHROW macro instead of noexcept for compatibility.

Teebonne:
Not sure this would fix the problem, or maybe I didn't understand what you mean with this macro.
A move operator should never throw exceptions (see [C++ Core Guidelines C.66](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rc-move-noexcept))

Teebonne:
Done, thanks!

gummif:
Thanks

PR #584, opened 2022-11-02
Rebuild with string_view argument, like the constructor

Crayon2000:
More convenient, like the constructor `message_t(std::string_view str)`

This is similar to PR #579 but with `std::string_view` instead of `std::string&`.

### DIFF
--- a/tests/message.cpp
+++ b/tests/message.cpp
@@ -179,6 +179,37 @@ TEST_CASE("message equality non equal lhs empty", "[message]")
     CHECK(msg_a != msg_b);
 }
 
+TEST_CASE("message rebuild with size", "[message]")
+{
+    const zmq::message_t msg();
+    msg.rebuild(5)
+    CHECK(msg.size() == 5);
+}
+
+#if defined(ZMQ_CPP11) && !defined(ZMQ_CPP11_PARTIAL)
+TEST_CASE("message rebuild with strings", "[message]")
+{
+    SECTION("string")
+    {
+        const std::string hi(data);
+        zmq::message_t hi_msg();
+        hi_msg.rebuild(hi)
+        CHECK(2u == hi_msg.size());
+        CHECK(0 == memcmp(data, hi_msg.data(), 2));
+    }
+#if CPPZMQ_HAS_STRING_VIEW
+    SECTION("string_view")
+    {
+        const std::string_view hi(data);
+        zmq::message_t hi_msg();
+        hi_msg.rebuild(hi)
+        CHECK(2u == hi_msg.size());
+        CHECK(0 == memcmp(data, hi_msg.data(), 2));
+    }
+#endif
+}
+#endif
+
 TEST_CASE("message to string", "[message]")
 {
     const zmq::message_t a;
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -544,6 +544,13 @@ class message_t
         rebuild(str.data(), str.size());
     }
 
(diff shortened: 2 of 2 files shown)


gummif:
Could you add a test case with string literal like `m.rebuild("hi")`?

tests/message.cpp
```diff
@@ -179,6 +179,45 @@ TEST_CASE("message equality non equal lhs empty", "[message]")
     CHECK(msg_a != msg_b);
 }
 
+TEST_CASE("message rebuild with size", "[message]")
+{
+    const zmq::message_t msg();
```
gummif, line 184: this should be non-const

tests/message.cpp
```diff
@@ -179,6 +179,45 @@ TEST_CASE("message equality non equal lhs empty", "[message]")
     CHECK(msg_a != msg_b);
 }
 
+TEST_CASE("message rebuild with size", "[message]")
+{
+    zmq::message_t msg();
```
gummif, line 184: You need to remove the () because of most vexing parse. Same for all default constructors below

Crayon2000:
Sorry for all the back and forth. I thought the CI would run automatically when the PR was submitted. I didn't expect you to be my compiler.

gummif:
Thats ok. For some reason I have to approve after each commit.

Crayon2000:
I'm a first-time contributor, so you need to approve the workflow.

Crayon2000:
@gummif, the test you asked to add with `hi_msg.rebuild("Hi")` is failing:
```
/home/runner/work/cppzmq/cppzmq/tests/message.cpp:193:24: error: call of overloaded ‘rebuild(const char [3])’ is ambiguous
  193 |     hi_msg.rebuild("Hi");
      |                        ^
In file included from /home/runner/work/cppzmq/cppzmq/tests/message.cpp:3:
/home/runner/work/cppzmq/cppzmq/zmq.hpp:542:10: note: candidate: ‘void zmq::message_t::rebuild(const string&)’
  542 |     void rebuild(const std::string &str)
      |          ^~~~~~~
/home/runner/work/cppzmq/cppzmq/zmq.hpp:548:10: note: candidate: ‘void zmq::message_t::rebuild(std::string_view)’
  548 |     void rebuild(std::string_view str)
      |          ^~~~~~~
```
The test works with the constructor because of this implementation:
```cpp
    // NOTE this constructor will include the null terminator
    // when called with a string literal.
    // An overload taking const char* can not be added because
    // it would be preferred over this function and break compatiblity.
    template<
      class Char,
      size_t N,
      typename = typename std::enable_if<detail::is_char_type<Char>::value>::type>
    ZMQ_DEPRECATED("from 4.7.0, use constructors taking iterators, (pointer, size) "
                   "or strings instead")
    explicit message_t(const Char (&data)[N]) :
        message_t(detail::ranges::begin(data), detail::ranges::end(data))
    {
    }
```
I don't think this a good idea to add this for `rebuild`.

So, should I remove the test for string literal? Or you want me to test my new implementation for string_view with `hi_msg.rebuild("Hi"sv)` (https://en.cppreference.com/w/cpp/string/basic_string_view/operator%22%22sv)? Or anything else?

gummif:
So what I was thinking is that currently there is only an overload for `std::string` so `rebuild("foo")` should work today. Then you added the overload for `string_view` so no there is an ambiguity and could break existing code. So I think the solution is to add another overload
```
void rebuild(const char* p)
{
  rebuild(p, std::strlen(p));
}
```
How does that sound?

PR #595, opened 2023-06-07
add function for adding file descriptor to poller_t

### LINKED ISSUE #557: poller_t and file descriptors (and why draft api)
Would it be possible to add the normal file descriptor into poller_t? It is possible in the c api of zmq so this functionality seems to be missing in cppzmq. 

I'm using cppzmq in a multithreaded program which merges measurement data from ADC cards. I have been mostly happy to cppzmq because of RAII. Now, I need to include slow control data into the program via UDP protocol. The sender is not in my hands and is not using zmq so I'm just using posix socket. I also would benefit to have the same poller been able to poll incoming zmq traffic and polling the normal file simultaneously. In meanwhile I have to use the c api for poller.

A side note. Is there any clear reason why the simple zmq::poller_t requires DRAFT api? This makes the installing of my software much more difficult since the installing of DRAFT support is not easy always.

uyha:
Fix #557

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -2744,8 +2744,8 @@ template<typename T = no_user_data> class poller_t
     void add_impl(fd_t fd, event_flags events, T *user_data)
     {
         if (0
-            != zmq_poller_add(poller_ptr.get(), fd, user_data,
-                              static_cast<short>(events))) {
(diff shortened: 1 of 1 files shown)


gummif:
Ok thanks, will fix that later.

PR #606, opened 2023-07-19
Add file descriptor support for poller

stephanschim:
Add file descriptor support for zmq::active_poller and zmq::base_poller_t.

Note: On Windows, only socket file descriptors are supported. This is a limitation of libzmq.

### DIFF
--- a/zmq_addon.hpp
+++ b/zmq_addon.hpp
@@ -34,8 +34,65 @@
 #include <limits>
 #include <functional>
 #include <unordered_map>
-#include <variant>
-#endif
+
+namespace zmq
+{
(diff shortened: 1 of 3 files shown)


uyha:
@gummif could you have a look at this PR?

zmq_addon.hpp
```diff
@@ -34,6 +34,7 @@
 #include <limits>
 #include <functional>
 #include <unordered_map>
+#include <variant>
```
gummif, line 37: This unfortunately requires C++17

gummif:
Looks good 👍

PR #608, opened 2023-07-20
Update CMakeLists.txt

LocutusOfBorg:
Require at least cmake 3.11 to run examples, as its required in main CMakeLists.txt file. Newer cmake issue errors with version less then 3.5 See e.g. https://bugs.debian.org/1040723


gummif:
Thanks

PR #612, opened 2023-08-25
Feature/expose monitor socket for active poller

dietelTiMaMi:
Motivation:
`zmq::monitor_t` in it's current form is inconvenient to use in combination with `zmq::active_poller_t`
the issue is, that the socket isn't available to the user of `zmq::monitor_t`

Proposal:
split `zmq::monitor_t::check_event()` to move the actual reading and processing of the events into a new protected method `zmq::monitor_t::process_event()`
provide a protected socket_ref for using the monitor_socket for polling in a derived class.

Execution:
The proposed changes were done to`zmq.hpp`
A test case demonstrating the desired usage was added to `tests/monitor.cpp`

By inheriting from `zmq::monitor_t` you can now implement custom ways of polling for socket events.
The modified api is hidden for "rookie" users that might accidentaly call `process_event()` without polling for changes before use.

### DIFF
--- a/tests/monitor.cpp
+++ b/tests/monitor.cpp
@@ -154,6 +154,8 @@ TEST_CASE("monitor from move assigned socket", "[monitor]")
 #if defined(ZMQ_BUILD_DRAFT_API) && defined(ZMQ_CPP11)                              \
   && !defined(ZMQ_CPP11_PARTIAL) && defined(ZMQ_HAVE_POLLER)
 #include "zmq_addon.hpp"
+#include <chrono>
+
 using namespace std::literals::chrono_literals;
 
 TEST_CASE("poll monitor events using active poller", "[monitor]")
--- a/tests/monitor.cpp
+++ b/tests/monitor.cpp
@@ -172,7 +172,7 @@ TEST_CASE("poll monitor events using active poller", "[monitor]")
         void addToPoller(zmq::active_poller_t &inActivePoller)
         {
             inActivePoller.add(
-              _monitor_socket, zmq::event_flags::pollin,
+              monitor_socket(), zmq::event_flags::pollin,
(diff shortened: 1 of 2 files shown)


dietelTiMaMi:
@gummif  Could you please relaunch the CI tests?
in the first run there was some C++14 code that didn't compile in C++11 inside the test
that is now fixed.

I also ran the CI in my fork and except for the Ubuntu 18.04 runners that are no longer existent everything went through.

dietelTiMaMi:
@gummif

Thanks for restarting the workflow.
Do you see any problems for merging this?
If you would like to have any changes, I will gladly contribute to be able to use it in our project.

gummif:
Hi I will try to take a look today.

tests/monitor.cpp
```diff
@@ -150,3 +150,89 @@ TEST_CASE("monitor from move assigned socket", "[monitor]")
...
+    addToPoller(sockets.server, poller);
+
+    sockets.init();
+    sockets.client.send(zmq::message_t(0), zmq::send_flags::dontwait);
+    CHECK(monitor.clientAccepted == 0);
+    CHECK(monitor.clientDisconnected == 0);
+
+    //Act
+    for (int i = 0; i < 10; i++) {
+        poller.wait(std::chrono::milliseconds(10));
```
gummif, line 221: could probably make more robust by increasing the wait time and breaking when monitor.clientAccepted > 0

gummif:
You decide if the comment is worth fixing, otherwise looks good.

tests/monitor.cpp
dietelTiMaMi: fixed according to your suggestion. In addition applied event filtering to monitor to only look for the desired events and ignore others.

PR #622, opened 2023-10-26
allow generic sequence of poller events

typon:
To avoid allocating (due to std::vector), I would like to use a static array for creating poller events. However, the `wait_all` method currently is hardcoded to accept only vectors. This PR addresses that by making `wait_all` a templated function.


gummif:
Nice, thanks.

PR #627, opened 2024-01-30
Add configurable sockopt for raw router sockets

dr7ana:
Since v3.3, libzmq has had the option for `ZMQ_ROUTER_RAW`, where [ZMQ framing is not added to read/write data](https://zguide.zeromq.org/docs/chapter2/#ZeroMQ-is-Not-a-Neutral-Carrier). This PR introduces the ability to set that using the already implemented logic as follows:

```C++
zmq::context_t ctx;
zmq::socket_t raw_router(ctx, zmq::socket_type::router);
raw_router.set(zmq::sockopt::router_raw, 1);
```


dr7ana:
@gummif, could you please OK the workflow when you get a chance?

dr7ana:
@gummif or @stephanlachnit, would you mind approving the workflow on this when you get a chance please?

PR #631, opened 2024-02-22
Add on_monitor_stopped function for monitor_t

egecetin:
Currently `ZMQ_EVENT_MONITOR_STOPPED` directly returns from function without an indicator. With a `on_monitor_stopped` function callback anyone can distinguish stop event from errors and might log this event.


egecetin:
@gummif Can you approve workflow?

PR #646, opened 2025-02-03
document CPPZMQ_BUILD_TESTS and adjust formatting and semantics of build instructions

nevadex:
document CPPZMQ_BUILD_TESTS
add git clone step
adjust semantics to be more clear
formatting


gummif:
Thanks, that's an improvement.

PR #652, opened 2025-03-12
Remove space in custom string literal

MichaelChirico:
I may lack sufficient context for this to be a useful PR, so please feel free to close if this is not welcome.

For no particular reason I saw an analogous PR to another R package repo in my GitHub home page:

https://github.com/marzer/tomlplusplus/pull/263

I searched the main R package repo for other similarly-affected packages, which includes {clustermq}, which bundles this repo as a submodule:

https://github.com/mschubert/clustermq/tree/master/src

Here's the documentation I could find about why this matters:

https://en.cppreference.com/w/cpp/language/user_literal

```c++
// error: all names that begin with underscore followed by uppercase
// letter are reserved (NOTE: a space between "" and _).
double operator"" _Z(long double);
```

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -1364,19 +1364,19 @@ constexpr const_buffer str_buffer(const Char (&data)[N]) noexcept
 
 namespace literals
 {
-constexpr const_buffer operator"" _zbuf(const char *str, size_t len) noexcept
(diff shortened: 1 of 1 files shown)


gummif:
We probably need to modify the clang-format file/update clang-format so this doesn't get reverted, but this is fine for now.

Issue #400, opened 2020-04-04
Bold plan for a major release


gummif:
Here are a few ideas for changes to be made for (potentially breaking) cppzmq v5.

1. Require C++11 standard support, but still support gcc 4.7 and later with partial C++11 support. This possible future requirement has been in the read me for 2 years now. Large fraction of the library now requires C++11.
2. Require libzmq 4.0+. There are no CI tests with older versions and there are no releases on github for version prior to 4.0.6. (should we maybe require 4.1 or 4.2?)
3. Remove deprecated functions that would not cause silent breakage (would cause compile time errors). Keep functions that would cause silent changes to runtime behavior if removed (e.g. greedy function overloads).

These changes are conservative in the sense that if you are using C++11 or later, libzmq v4 or later and not using any deprecated functionality, then the upgrade from the latest v4 to v5 should not require any changes to you code. If you do use deprecated functions then you might get a compile time error in the worst case.

bluca:
Please do not break backward compatibility. This header is shipped in distributions. It would cause major, major headaches and work to downstreams.

gummif:
After some deep though I can see that part 3 and maybe part 2 could cause some pain with system packages (system packages are generally a bad idea anyway for development). I don't know how the process is for adding new package versions to distributions, but I guess it is their responsibility to not add package versions that would break the system? But if it is possible people will eventually managed to break things.

About requiring C++11, the read me says "Some features already require a compiler supporting C++11. In the future, probably all features will require C++11". Is that something that you prefer never happens?

And regarding part2, I believe libzmq v3.x is rarely used. And it would be very unlikely that someone would use libzmq v3.x together with latest cppzmq, so in practice that would requirement should have no effect (in that event an option would be to simply update libzmq or not update cppzmq). Correct me if you can come up with a situation where that would be the case.

bluca:
> I don't know how the process is for adding new package versions to distributions, but I guess it is their responsibility to not add package versions that would break the system?

Yes, but it means there will never be a new version uploaded, which is not good.

The problem is not C++11 so much for distros - they all support it, it's just the ancient and deprecated stuff that doesn't. Not even zmq 3.x. The problem is breaking API compatibility by removing functions.

bluca:
That would be fine, yes

planetmarshall:
> Here are a few ideas for changes to be made for (potentially breaking) cppzmq v5.
> 
[...]

PR #653, opened 2025-04-08
[RFC] Add initial C++20 coroutine support

SidneyCogdill:
The initial C++20 coroutine support. I may implement C++26 sender receiver support later, once I fully grasped its design (ref: [stdexec](https://github.com/NVIDIA/stdexec)).

The async version of the socket is placed under `zmq::async::corral`. It's implemented for [corral](https://github.com/hudson-trading/corral) but any open coroutine implementation can `co_await` on the `send` / `recv` operations. [Boost.Cobalt](https://www.boost.io/library/latest/cobalt/) is another open coroutine implementation.

Issues:

1. The multipart send/receive cancellation isn't covered in the test cases.
2. Relies on Boost.Asio. the standalone version of Asio isn't supported yet (but should be trivial to do so; I'm just tired currently and pushing the work up until now)
3. Cancellation only works with corral currently because C++20 coroutine doesn't have a standardized way to do cancellation (Note that C++26 sender receiver has standardized cancellation support)
4. Many code duplication, because clangd doesn't give helpful auto completions within function templates.
5. It's strictly single thread concurrency as it's written for corral. 1:N multiplexing can be safely done by running multiple `io_context` in parallel. M:N multiplexing isn't possible currently with corral (Note that C++26 sender receiver does have proper M:N multiplexing)
6. I didn't test it on Windows, only on Linux. It should work but I don't know.
7. The comments are largely incomplete. I will write them properly in a follow-up commit.
8. No documentations or examples. Again will do them in follow-up commits.

Alternatively wait for my implementation for sender receiver support (again, once I've grasped it myself), which should resolve 3 and 5.

Also refer to https://github.com/zeromq/cppzmq/issues/400

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "tests/third_party/corral"]
+	path = tests/third_party/corral
+	url = https://github.com/hudson-trading/corral
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -52,6 +52,7 @@ if (CPPZMQ_TEST_COROUTINE)
     target_sources(
       unit_tests PRIVATE
       async/corral/message.cpp
+      async/corral/common.hpp
     )
 endif()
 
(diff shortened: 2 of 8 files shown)


gummif:
Thanks for this. I guess the sender approach is superior, so this can be closed?

PR #656, opened 2025-04-17
Fix std::max compilation error on Windows

### LINKED ISSUE #654: `std::max` causes compilation error on Windows
There is one use of `std::max` that doesn't use the `(std::max)` trick when the `max` macro is defined by Windows.

https://github.com/zeromq/cppzmq/blob/21c83ca0c3d8c11d81aa13aca2a565607a25920b/zmq.hpp#L2756

My current workaround is to add `#define NOMINMAX` to all my files but it would be nice if the library could handle it.

cacharle:
Fixes #654


cacharle:
I might be blind but I do not see the Windows CI/CD compile errors

gummif:
Yeah, it has not worked for a while. I think this is fine since it works on gcc and clang.

PR #657, opened 2025-04-17
Add zmq_timers support

### LINKED ISSUE #575: Support for zmq timers?
Hi,

looking at the library's source, I guess there is no support for timer functions (the ones described in http://api.zeromq.org/master:zmq-timers).
Is there a safe way to use zmq timers while continuing to use cppzmq for networking?

Are there any plans to support them?

Thanks!

cacharle:
Fixes #575 

This is a draft to add support for the `zmq_timers_*` functions.

Please let me know what are the requirements to get it merged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 3.11)
 
 list (APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 include (DetectCPPZMQVersion)
 
--- a/tests/timers.cpp
+++ b/tests/timers.cpp
@@ -1,0 +1,80 @@
+#include <catch2/catch_all.hpp>
+#include <zmq.hpp>
+
+#include <type_traits>
+#include <thread>
+#include <chrono>
+
+#if defined(ZMQ_CPP11) && defined(ZMQ_HAVE_TIMERS)
+
+static_assert(std::is_default_constructible<zmq::timers>::value);
+static_assert(!std::is_copy_constructible<zmq::timers>::value);
+static_assert(!std::is_copy_assignable<zmq::timers>::value);
+
+TEST_CASE("timers constructor", "[timers]")
+{
+    zmq::timers timers;
+    CHECK(!timers.timeout().has_value());
+}
+
+TEST_CASE("timers add/execute", "[timers]")
+{
+    using namespace std::chrono_literals;
+    zmq::timers timers;
+    bool handler_ran = false;
+    timers.add(4ms, [](auto, void *arg) { *(bool *) arg = true; }, &handler_ran);
+    CHECK(timers.timeout().has_value());
+    CHECK(!handler_ran);
+    std::this_thread::sleep_for(10ms);
+    timers.execute();
+    CHECK(handler_ran);
+}
+
+TEST_CASE("timers add/cancel", "[timers]")
+{
+    using namespace std::chrono_literals;
+    zmq::timers timers;
+    bool handler_ran = false;
+    auto id =
+      timers.add(4ms, [](auto, void *arg) { *(bool *) arg = true; }, &handler_ran);
+    CHECK(timers.timeout().has_value());
+    CHECK(!handler_ran);
+    timers.cancel(id);
+    CHECK(!timers.timeout().has_value());
+    CHECK(!handler_ran);
+}
+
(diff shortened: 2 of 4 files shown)


zmq.hpp
```diff
@@ -2794,6 +2794,70 @@ inline std::ostream &operator<<(std::ostream &os, const message_t &msg)
...
+class timers_t
+{
+  public:
+    timers_t() : _timers(zmq_timers_new())
+    {
+        if (_timers == nullptr)
+            throw error_t();
+    }
+
+    timers_t(timers_t &other) = delete;
```
gummif, line 2808:
```suggestion
    timers_t(const timers_t &other) = delete;
```

gummif: Also need to delete the copy assignment operator.

zmq.hpp
```diff
@@ -2794,6 +2794,70 @@ inline std::ostream &operator<<(std::ostream &os, const message_t &msg)
...
+
+    ~timers_t() { ZMQ_ASSERT(zmq_timers_destroy(&_timers) == 0); }
```
gummif, line 2810: Move zmq_timers_destroy out of the assert, and only check the return value in the assert.

zmq.hpp
```diff
@@ -2794,6 +2794,70 @@ inline std::ostream &operator<<(std::ostream &os, const message_t &msg)
     return os << msg.str();
 }
 
+using timer_id_t = int;
```
gummif, line 2797: This whole diff should be behind a preprocessor if, checking the min required libzmq version. Also requires C++11, but it uses `std::optional` so either require C++17 or use the `trivial_optional` for pre C++17 version (but not sure if `milliseconds` is trivial). Otherwise just require C++17.

gummif: Maybe add `using timer_fn_t = zmq_timer_fn;`

gummif (changes requested):
Thanks. What is missing is some minor fixes and some simple tests should be sufficient.

cacharle:
I've had to rename `zmq::timers_t` to `zmq::timers`, the first one was making the tests crash (I'm guessing a linker name conflict with libzmq)

Do you have any workaround to keep the `_t` convention?

gummif:
I think `zmq::timers` is fine. Not a fan of the _t convention, also unfortunate we are using the same namespace as libzmq.

zmq.hpp
gummif: I think milliseconds is fine as is. But duration would be an improvement.

cacharle:
Alright then I think this PR is ready

zmq.hpp
```diff
@@ -2794,6 +2794,85 @@ inline std::ostream &operator<<(std::ostream &os, const message_t &msg)
...
+        int rc = zmq_timers_reset(_timers, timer_id);
+        if (rc == -1)
+            throw zmq::error_t();
+    }
+
+    timeout_result_t timeout() const
+    {
+        int timeout = zmq_timers_timeout(_timers);
+        if (timeout == -1)
+            return std::nullopt;
```
gummif, line 2859: This needs to be `timeout_result_t{}` instead.

tests/timers.cpp
```diff
@@ -0,0 +1,78 @@
+#include <catch2/catch_all.hpp>
+#include <zmq.hpp>
+
+#include <type_traits>
+#include <thread>
+#include <chrono>
+
+#ifdef ZMQ_CPP11
```
gummif, line 8: You should guard all all this file with `#if defined(ZMQ_CPP11) && defined(ZMQ_HAVE_TIMERS)`

CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 3.11)
 
 list (APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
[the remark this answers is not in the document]
cacharle: don't most editors with LSP use build/compile_commands.json? I thought it might be handy for someone else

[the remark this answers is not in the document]
gummif: Not sure, does this affect the consumer of the library? But it can stay if it is useful.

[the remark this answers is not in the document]
cacharle: https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html

It only generates a `compile_commands.json` file

gummif:
Could you try rebasing your branch on latest master, I just fixed the CI runners.

gummif:
Very good. Thanks for this.

PR #665, opened 2025-09-18
Update zmq::message_t::str with max_size

### LINKED ISSUE #663: Update the zmq::messsage_t::str() method to allow to print larger messages
There is an arbitrary limitation of 1000 size to print the `[too big to print]` placeholder:

https://github.com/zeromq/cppzmq/blob/3bcbd9dad2f57180aacd4b4aea292a74f0de7ef4/zmq.hpp#L707C9-L707C28

It would be nice to have this limit as an argument instead:

```cpp
    std::string str(size_t max_size = 1000) const
```

### LINKED ISSUE #664: Update zmq::multipart_t::send to support zmq::send_flags
https://github.com/zeromq/cppzmq/blob/3bcbd9dad2f57180aacd4b4aea292a74f0de7ef4/zmq_addon.hpp#L474

`zmq::multipart_t::send` still uses the old `int` send flags.

cacharle:
Fixes #663

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -691,39 +691,43 @@ class message_t
     *   Use to_string() or to_string_view() for
     *   interpreting the message as a string.
     */
-    std::string str() const
+    std::string str(size_t max_size = 1000) const
     {
         // Partly mutuated from the same method in zmq::multipart_t
         std::stringstream os;
 
         const unsigned char *msg_data = this->data<unsigned char>();
         unsigned char byte;
-        size_t size = this->size();
+        size_t size_to_print = std::min(this->size(), max_size);
         int is_ascii[2] = {0, 0};
+        // Set is_ascii for the first character
+        if (size_to_print > 0) {
+            byte = *msg_data;
+            is_ascii[0] = (byte >= 32 && byte < 127);
+        }
 
         os << "zmq::message_t [size " << std::dec << std::setw(3)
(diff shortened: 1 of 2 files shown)


zmq.hpp
```diff
@@ -691,39 +691,43 @@ class message_t
...
-    std::string str() const
+    std::string str(size_t max_size = 1000) const
     {
         // Partly mutuated from the same method in zmq::multipart_t
         std::stringstream os;
 
         const unsigned char *msg_data = this->data<unsigned char>();
         unsigned char byte;
-        size_t size = this->size();
+        size_t size_to_print = std::min(this->size(), max_size);
```
gummif, line 701: while at it, make this const

zmq.hpp
```diff
@@ -691,39 +691,43 @@ class message_t
...
-                if (is_ascii[1]) {
-                    os << byte;
-                } else {
-                    os << std::hex << std::uppercase << std::setw(2)
-                       << std::setfill('0') << static_cast<short>(byte);
-                }
-                is_ascii[0] = is_ascii[1];
+           << std::setfill('0') << this->size() << "] (";
+        while (size_to_print--) {
+            byte = *msg_data++;
```
gummif, line 712: `const unsigned char byte = ...` and remove `unsigned char byte;` above.

PR #683, opened 2026-03-20
fix: rename parameter to avoid -Wshadow warning in message_t constructor

### LINKED ISSUE #641: Parameter name "data" in message_t constructor reuses the method name (warning when compiling user code)
gcc issues a warning:
```
warning: declaration of 'data' shadows a member of 'zmq::message_t' [-Wshadow]
```
It's issued by this constructor declaration:

```
    template<
      class Char,
      size_t N,
      typename = typename std::enable_if<detail::is_char_type<Char>::value>::type>
    ZMQ_DEPRECATED("from 4.7.0, use constructors taking iterators, (pointer, size) "
                   "or strings instead")
    explicit message_t(const Char (&data)[N]) :
        message_t(detail::ranges::begin(data), detail::ranges::end(data))
    {
    }

    template<class Range,
             typename = typename std::enable_if<
               detail::is_range<Range>::value
               && ZMQ_IS_TRIVIALLY_COPYABLE(detail::range_value_t<Range>)
               && !detail::is_char_type<detail::range_value_t<Range>>::value
               && !std::is_same<Range, message_t>::value>::type>
    explicit message_t(const Range &rng) :
        message_t(detail::ranges::begin(rng), detail::ranges::end(rng))
    {
    }
```

It actually doesn't matter if this is deprecated - this problem is reported regardless whether this constructor is in use or not.

It's easy to fix - just use a different name than `data` - which is simultaneously a name of the method in `message_t`.

ghost:
## Summary

Renames the `data` parameter in the deprecated array constructor of `message_t` to `arr` to avoid GCC's `-Wshadow` warning:

```
warning: declaration of 'data' shadows a member of 'zmq::message_t' [-Wshadow]
```

The parameter `data` shadows the `message_t::data()` method. Renaming to `arr` resolves the warning without changing behavior (the constructor is already deprecated).

Fixes #641

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -483,8 +483,8 @@ class message_t
       typename = typename std::enable_if<detail::is_char_type<Char>::value>::type>
     ZMQ_DEPRECATED("from 4.7.0, use constructors taking iterators, (pointer, size) "
                    "or strings instead")
-    explicit message_t(const Char (&data)[N]) :
-        message_t(detail::ranges::begin(data), detail::ranges::end(data))
+    explicit message_t(const Char (&arr)[N]) :
+        message_t(detail::ranges::begin(arr), detail::ranges::end(arr))
     {
     }
 


gummif:
Thanks

PR #684, opened 2026-03-26
Improve constexpr coverage in zmq.hpp with C++11/14/17 compatibility

mberk-yilmaz:
## What changed

This PR expands `constexpr` usage in `zmq.hpp` with a compatibility-first approach across C++11/14/17.

### `zmq.hpp`
- Define `ZMQ_EXTENDED_CONSTEXPR` as `constexpr` (under existing compiler/version guards).
- Add/extend `constexpr` on safe pointer-state observers:
  - `context_t::operator bool() const`
  - `detail::socket_base::handle() const`
  - `detail::socket_base::operator bool() const`
  - `detail::socket_base::connected() const`
- Keep non-const observer forms conditionally `constexpr` only when `ZMQ_EXTENDED_CONSTEXPR` is available.
- Make `socket_ref` null-comparison operators `constexpr`.
- Make `detail::socket_base` `operator==`/`operator!=` `constexpr` and use direct pointer equality.
- Add constexpr-friendly updates for trivial value wrappers:
  - `recv_buffer_size::truncated()`
  - `detail::trivial_optional<T>` value constructor, `operator bool()`, `has_value()`.

### Tests
- Strengthen compile-time coverage with `static_assert`:
  - `tests/socket.cpp`: `socket_base` constexpr handle/bool checks
  - `tests/socket_ref.cpp`: constexpr null comparison checks
  - `tests/context.cpp`: constexpr checks for `recv_buffer_size::truncated()` and fallback `trivial_optional` (`!CPPZMQ_HAS_OPTIONAL`)

## Why this is safe

- No runtime ZMQ call paths were made `constexpr`; changes are limited to trivial observers/value-state helpers.
- Existing compiler guards are preserved; `extended constexpr` remains opt-in by language/compiler capability.
- No ABI-impacting layout changes.
- Null/pointer semantics are unchanged (only compile-time evaluability improved).
- Potentially risky areas (buffer pointer arithmetic, pointer ordering operators) were intentionally left unchanged for portability/semantic safety.

## Compatibility notes

- C++11: baseline support preserved.
- C++14/17: additional constexpr opportunities activate via existing guard logic.
- Legacy compiler constraints already encoded in `ZMQ_EXTENDED_CONSTEXPR` remain respected.

## Validation

Validated by building and testing in three standards:
- C++11: build + full unit tests pass
- C++14: build + full unit tests pass
- C++17: build + full unit tests pass

### DIFF
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -985,7 +985,7 @@ template<class T> class trivial_optional
     static_assert(std::is_trivial<T>::value, "T must be trivial");
     using value_type = T;
 
-    trivial_optional() = default;
+    ZMQ_CONSTEXPR_FN trivial_optional() = default;
     ZMQ_CONSTEXPR_FN trivial_optional(T value) noexcept :
         _value(value), _has_value(true)
     {
@@ -2337,11 +2337,7 @@ class socket_t : public detail::socket_base
 
     ~socket_t() ZMQ_NOTHROW { close(); }
 
-#ifdef ZMQ_EXTENDED_CONSTEXPR
     ZMQ_EXTENDED_CONSTEXPR operator void *() ZMQ_NOTHROW { return _handle; }
-#else
-    operator void *() ZMQ_NOTHROW { return _handle; }
-#endif
 
     ZMQ_CONSTEXPR_FN operator void const *() const ZMQ_NOTHROW { return _handle; }
 
--- a/zmq.hpp
+++ b/zmq.hpp
@@ -752,11 +752,17 @@ class message_t
     }
 
 #ifdef ZMQ_EXTENDED_CONSTEXPR
-    ZMQ_NODISCARD ZMQ_EXTENDED_CONSTEXPR zmq_msg_t *handle() ZMQ_NOTHROW { return &msg; }
+    ZMQ_NODISCARD ZMQ_EXTENDED_CONSTEXPR zmq_msg_t *handle() ZMQ_NOTHROW
+    {
+        return &msg;
+    }
 #else
     ZMQ_NODISCARD zmq_msg_t *handle() ZMQ_NOTHROW { return &msg; }
 #endif
-    ZMQ_NODISCARD ZMQ_CONSTEXPR_FN const zmq_msg_t *handle() const ZMQ_NOTHROW { return &msg; }
+    ZMQ_NODISCARD ZMQ_CONSTEXPR_FN const zmq_msg_t *handle() const ZMQ_NOTHROW
+    {
+        return &msg;
+    }
 
   private:
     //  The underlying message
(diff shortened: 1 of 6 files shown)


gummif:
Why these two ways of marking functions constexpr?
```c++
// why not just use ZMQ_CONSTEXPR_FN?
// or if there is need why not just ZMQ_EXTENDED_CONSTEXPR operator void *() ZMQ_NOTHROW { return _handle; }
#ifdef ZMQ_EXTENDED_CONSTEXPR
    ZMQ_EXTENDED_CONSTEXPR operator void *() ZMQ_NOTHROW { return _handle; }
#else
    operator void *() ZMQ_NOTHROW { return _handle; }
#endif

    ZMQ_CONSTEXPR_FN operator void const *() const ZMQ_NOTHROW { return _handle; }
```

Related to this it's interesting to still support C++98 with C++26 releasing this year.

zmq.hpp
```diff
@@ -965,7 +986,10 @@ template<class T> class trivial_optional
     using value_type = T;
 
     trivial_optional() = default;
```
gummif: ZMQ_CONSTEXPR_FN also?

zmq.hpp
```diff
@@ -2164,42 +2214,48 @@ ZMQ_CONSTEXPR_VAR from_handle_t from_handle =
...
     return !(sr == nullptr);
 }
 #endif
 
-inline bool operator==(const detail::socket_base &a,
-                       const detail::socket_base &b) ZMQ_NOTHROW
+inline ZMQ_CONSTEXPR_FN bool operator==(const detail::socket_base &a,
+                                        const detail::socket_base &b) ZMQ_NOTHROW
 {
-    return std::equal_to<const void *>()(a.handle(), b.handle());
```
gummif, line 2199: keep this to avoid UB

mberk-yilmaz: `std::equal_to<T>::operator()` is `constexpr` only since C++14.
So in C++11 mode it cannot be used in a `ZMQ_CONSTEXPR_FN` function body.
Also, pointer equality (`==`) is well-defined and not UB. Why do you think it's UB?

zmq.hpp
mberk-yilmaz: For intent, yes, for behavior there is no difference.

As cppreference states:
"If it satisfies the requirements of a constexpr constructor (until C++23) / constexpr function (since C++23), the generated constructor is constexpr."

zmq.hpp
gummif: Ok you are correct this is fine. The confusion is that comparing possibly unrelated pointers with < is unspecified unless using `std::less`.

zmq.hpp
```diff
@@ -751,8 +751,18 @@ class message_t
         std::swap(msg, other.msg);
     }
 
+#ifdef ZMQ_EXTENDED_CONSTEXPR
```
gummif, line 754: All these #ifdef ZMQ_EXTENDED_CONSTEXPR branches can be removed

PR #687, opened 2026-05-24
feat: add send_flags overload for multipart_t::send

### LINKED ISSUE #664: Update zmq::multipart_t::send to support zmq::send_flags

granite-kas:
# Update zmq::multipart_t::send to support zmq::send_flags

Fixes zeromq/cppzmq#664.

## Summary

- Add a `send_flags` overload for `zmq::multipart_t::send`.
- Keep the existing integer-based overload for backward compatibility.
- Forward the new overload to the existing implementation so current behavior stays unchanged.

## Notes

This is a small API improvement in the header-only `zmq_addon.hpp` file. It makes multipart send calls consistent with the rest of the modern cppzmq API.

### DIFF
--- a/zmq_addon.hpp
+++ b/zmq_addon.hpp
@@ -491,6 +491,13 @@ class multipart_t
         return true;
     }
 
+#ifdef ZMQ_CPP11
+    bool send(socket_ref socket, send_flags flags)
+    {
+        return send(socket, static_cast<int>(flags));
+    }
+#endif
+
     // Concatenate other multipart to front
     void prepend(multipart_t &&other)
     {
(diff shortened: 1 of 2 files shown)


zmq_addon.hpp
```diff
@@ -381,7 +381,12 @@ class multipart_t
 
     typedef std::deque<message_t>::reverse_iterator reverse_iterator;
     typedef std::deque<message_t>::const_reverse_iterator const_reverse_iterator;
-
+    #ifdef ZMQ_CPP11
+    bool send(socket_ref socket, send_flags flags = send_flags::none)
```
gummif: Can you move this below the other send definition, and remove the default ` = send_flags::none`.

granite-kas: Thanks! I moved the send_flags overload below the existing int-based send overload and removed the default argument.

